PR #26, opened 2018-01-10
Create a new test container to run libiscsi - iSCSI compliance test suite 

ksatchit:
Changes: 
----------

- libiscsi is a popular opensource iSCSI test suite. It checks for both generic SCSI and iSCSI compliance on the exposed iSCSI target (https://github.com/sahlberg/libiscsi)

- As part of this PR, the libiscsi tool is containerized, with a generic script that runs the test suite and logs results

- It is recommended to attach a host dir as a bind-mount into the container during run so that the logs can be persisted. Or better, use PVs like openebs

- A kubernetes job which uses this libiscsi container will be created which will enable us to run "compliance" tests on any storage (block). 

Tested On: 
-------------
- Container tested after building locally on ubuntu 16.04 64 bit Baremetal boxes/ESX VM


libiscsi/Dockerfile
```diff
@@ -0,0 +1,6 @@
+FROM ubuntu
```
yudaykiran: Can this be checked if it runs with alpine? Just to reduce the size of the image, in case it works.

ksatchit: libiscsi is not built for alpine and the effort to get it running (install all dependencies during compile)  is considerable. Since this is a compliance suite - the expectation is that it is a one-time run should any "user" attempt to use it. 

Having said that, agree on the intent. Have replaced ubuntu w/ debian which I found to be next best wrt size and also having support for the libiscsi package. It is ~35 MB lesser.

The tests were successfully run w/ the updated dockerfile.

PR #31, opened 2018-04-25
Better organize the tests within Litmus repo

kmova:
This project has been hosting source code of several test containers that were being used to test the openebs (as part of its e2e). Some of these tests in e2e are very generic and can actually be used against any storage provider within kubernetes. Taking this a step further if this project can host the actual tests, it would benefit both the users and other storage solutions to test against Litmus by modifying the Storage Class associated with the workload. 

To help with this requirement, the code has been re-structured to push all the test-container source code under "tools" directory. All the generic workload related tests will be pushed to "tests" folder, while orchestration these tests across different environments will be taken care by the code in "executor" folder. Currently, the code supports using Ansilbe as executor. 

**Notes to the reviewer**

The amount of files changed is very little, however the movement involved a lot of files. Also in the process of this movement, I have:
- removed the k8s-client from this repository. 
- removed the automated builds on the test containers, which anyway was not required.


README.md
```diff
@@ -1,39 +1,107 @@
...
+```
+git clone https://github.com/openebs/litmus.git
+cd litmus/executor/ansible
+<modify/customize>
+run-litmus.sh
+```
+
+The above script will verify that it has all the details required for it to proceed. And it provides you with a helpful message on the progress of the tests.
+
+*Litmus may take a while to show a reaction as it puts the system through rigorous scrutiny!*
```
ksatchit: :smile:

PR #55, opened 2018-05-26
Used `with` construct to handle files open/close.

shubhambhattar:
Also, changed shebang line to let `env` find `python3` executable.  

Signed-off-by: Shubham Bhattar <shubhambhattar@gmail.com>

### DIFF
--- a/executor/ansible/roles/inventory/files/inventory.py
+++ b/executor/ansible/roles/inventory/files/inventory.py
@@ -22,9 +22,8 @@ class Inventory(object):
                     if c == 1:
                         hostaliasname = '%s0%s ansible_ssh_host' % (i[0], c)
                         config[hostgroupname] = {hostaliasname: i[1]}
-                        f = open(path, 'w')
-                        config.write(f)
-                        f.close()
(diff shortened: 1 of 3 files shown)


shubhambhattar:
@muratkars I don't understand this. Why is the DCO expected format different here? Instead of `name@email.com`, now I have to enter `name@users.noreply.github.com`. I don't understand how do I know that I need to enter the latter one here?

shubhambhattar:
@ksatchit What should I do?

ksatchit:
@shubhambhattar , thanks for the PR. Was trying to understand this - The DCO expects the "commit author email" as part of the check and this is generally in the form of `username@users.noreply.github.com` by default if the `Keep my email address private` setting is enabled. Could you please check if there is a mismatch in your Sign-off email and the commit-author email ?

shubhambhattar:
@ksatchit Yes, I have checked the box that says `Keep my email address private`.

shubhambhattar:
@ksatchit Check [#56](https://github.com/openebs/litmus/pull/56).

PR #56, opened 2018-05-29
 Used `with` construct to handle files open/close. 

shubhambhattar:
Also, changed shebang line to let `env` find `python3` executable.

Signed-off-by: Shubham Bhattar <shubhambhattar@users.noreply.github.com>


shubhambhattar:
@ksatchit Can you add the `summerhack` label here?

PR #78, opened 2018-07-31
Add playbbok for create & delete pre-req. in AWS

chandankumar4:

- Add `pre-requisite.yml`, it will create all dependencies for creating cluster in AWS like (vpc, subnet, Internet-gateway and route-table).

- Add `delete-pre-requisite.yml`, it will delete all dependencies created by `pre-requisite.yml` after deletion of cluster in AWS.

Signed-off-by: Chandan Kumar <chandan.kr404@gmail.com>

### DIFF
--- a/k8s/aws/k8s-installer/delete-pre-requisite.yml
+++ b/k8s/aws/k8s-installer/delete-pre-requisite.yml
@@ -1,0 +1,63 @@
+# delete-pre-requisite.yml
+# Description:  Deleting pre requisite will delete VPC, Subnet and Internet Gateway
+
+###############################################################################################
+#Test Steps:
+
+#1. Disassociating route table fro aws.
+#2. Delete route table from aws.
+#3. Detach internet gateway from aws.
+#4. Delete internet gateway from aws.
+#4. Deleting the subnet from aws.
+#5. Deleting the VPC Network form aws.
+#6. Deleting temporary file which store the VpcId and SubnetId
+
+###############################################################################################
+
+---
+
+- hosts: localhost
+
+  vars:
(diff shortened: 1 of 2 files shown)


ksatchit:
Hi @chandankumar4 , firstly, good work on the playbooks to setup the pre-requisites for creation on AWS cluster!! Approving this PR for merge for now, with some recommendations for subsequent improvements/enhancements.

- There are ansible modules available for most/all the steps (VPC, subnet, internet gateway, route-table creation etc.,).  The shell modules using aws-cli can be replaced with these. Ansible modules are idempotent in nature (whereas shell is not) and will help in case of re-runs after failure of one of the steps if needed. They are also more readable/understandable (https://docs.ansible.com/ansible/2.5/modules/list_of_cloud_modules.html#amazon) 

- We can add confirmation checks or failure conditions on the individual steps so that playbooks fails early. 

- Interested to know rollback approach upon failure. The delete pre requisites may need to handle incomplete setup.

- We can use folded scalars ">" or "|" operators in the shell module in case of multi-line commands

chandankumar4:
Hi @ksatchit thanks for reviewing!
- For improvements/enhancements i'm looking into it i.e for replacing the shell module.

- For rollback approach, If `pre-requisite.yml` fail then we have to run `delete-pre-requisite.yml` which contains `ignore_errors: True` i.e if present then delete otherwise move forward and delete others.
     I will automate this also like if `pre-requisite.yml` fail then, in clean up process it automatically call `delete-pre-requisite.yml` and clean everyting.

ksatchit:
Thanks @chandankumar4 , could you provide a playbook run sample if possible.

chandankumar4:
sure! @ksatchit

chandankumar4:
pre-requisite
delete-pre-requisite

PR #94, opened 2018-08-12
Add litmus job to install storageclass for v0.6

ashishranjan738:
Signed-off-by: Ashish Ranjan <ashishranjan738@gmail.com>

-> This PR adds litmus job to create storage class and storage pool for openebs v0.6.
-> By default this job creates a storage pool with name *openebs-mntdir* and storage class with name *openebs-storageclass*
-> storage class and storage pool information can be overided by providing it explicitly through environment variable.
-> Environment variables that can be used :

- STORAGE_POOL_NAME  
- STORAGE_PATH  
- STORAGECLASS_NAME  
- JIVA_REPLICA_COUNT  
- CAPACITY
- VOLUME_MONITOR

-> This PR has been tested on GCP.


providers/openebs/installers/storageclass/0.6/ansible/storageclass.yaml
```diff
@@ -0,0 +1,68 @@
...
+          - name: Applying storage pool
+            shell: "{{ kubeapply }} apply -f {{ storage_pool }}" 
+            args:
+              executable: /bin/bash
+              
+          - name: Applying storage class
+            shell: "{{ kubeapply }} apply -f {{ storage_class }}" 
+            args:
+              executable: /bin/bash
+
```
ksatchit: Can we add a couple of simple checks to validate sp & sc installation

providers/openebs/installers/storageclass/0.6/ansible/storageclass_setup.yaml
```diff
@@ -0,0 +1,54 @@
...
+
+          - template:
+              src: sc.j2
+              dest: "{{ storage_class }}"
+
+          - name: Applying storage pool
+            shell: "{{ kubeapply }} apply -f {{ storage_pool }}" 
+            args:
+              executable: /bin/bash
+            ignore_errors: True
```
ksatchit: @ashishranjan738 , any particular reason for adding the ignore_errors (are you possibly factoring in existing pools/sc with the same names? )

ksatchit: Got it. We should be OK for actual deploy failures as the job will fail due to validations for the resources in subsequent tasks. However, thinking about this case: 

- It is possible that the CI has passed custom values for the params via ENV (diskpaths, replica counts etc.,), which may/may-not be present in the existing SP/SC. That way, the application deployment job will be affected. 

What do you think about handling it by removing the existing class and launching newly. Even if there are existing application PVCs using the same storage class, it should be OK as the pods will be already deployed.

ashishranjan738: ok so let's stop our job as a failure when we find any previous sc or sp artifacts.

ashishranjan738:
@ksatchit made the changes as requested

PR #103, opened 2018-08-16
Update tests to generate unique test-specific job names

ksatchit:

- Currently, all tests have job name as "litmus"  (The tests cleanup all the resources created as part of the test, with the litmus job itself stays behind). 

  - The job does not identify with the test being performed
  - Re-running these tests errors out & can be executed only if the completed job instance is removed/deleted. It is possible for users to forget this step. 
  - CI systems using litmus jobs in their pipelines will have the additional step of deletion of completed job in each stage, whereas a bulk removal at the end of the pipeline may be more desirable. 

- This PR addresses the above issue by : 

  a) Prefixing test names to litmus test jobs (`litmus-<test-name>`)
  b) Adding metadata.generateName as against metadata.name to make the name unique (hash appended to (a)

- The above changes mandate that the litmus job should be deployed via `kubectl create -f` as against `kubectl apply`. As apply doesn't support the generateName API

- The job removal should be performed via `kubectl delete job <job>` as against a `delete -f <spec>` as the latter cannot resolve/find the unique name

### DIFF
--- a/README.md
+++ b/README.md
@@ -58,7 +58,7 @@ constituting a separate test. Select a workload and follow the instructions unde
   For example, to run a MySQL benchmarking test:
 
 ```
-cd tests/mysql/mysql_storage_benchmark/
+cd apps/percona/tests/mysql_storage_benchmark/
 <Modify the PROVIDER_STORAGE_CLASS in run_litmus_test.yaml>
 kubectl create -f run_litmus_test.yaml
 ```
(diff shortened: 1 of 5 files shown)


apps/percona/tests/mysql_data_persistence/run_litmus_test.yaml
```diff
@@ -2,12 +2,11 @@
 apiVersion: batch/v1
 kind: Job
 metadata:
-  name: litmus
+  generateName: litmus-mysql-data-persistence-
```
kmova: :+1:

README.md
```diff
@@ -60,7 +60,7 @@ constituting a separate test. Select a workload and follow the instructions unde
 ```
 cd tests/mysql/mysql_storage_benchmark/
```
kmova: The directory needs to be updated..

ksatchit: Done

PR #107, opened 2018-08-16
 Add timeout functionality to terminate playbook after 900s

harshvkarn:
[Enhancement]
- Add Multiprocessing to enable timeout feature
- Program exits with code 0 when Cluster is up and Running, else returns
  1 as exit code after 900 seconds of checking
- Remove Block and Rescue to terminate script with error, when any test fails

Earlier script was using `signal`, a python library to terminate after timeout. which was not suffice and was naive appraoch. Currently it is using `multiprocessing` to achieve the same, but more effectively.


harshvkarn:
@ksatchit block and rescue were handling any error and returns the execution as successful termination even the playbook fails. The CI tools, like Jenkins and Gitlab becomes oblivion of `Test Failed` and were always passing even in the case of task failure in playbook.

ksatchit:
@harshvkarn , ideally, the gitlab job should parse for failure flags returned by the playbook and appropriately mark the job/stage as failed. Now, there is also a provision to use result Custom Resource that can be updated with test/playbook result. For now, I am going ahead with merge - let us track this requirement with an issue and raise a separate PR for the same.

harshvkarn:
sure @ksatchit

PR #109, opened 2018-08-18
Add litmus chaos jobs for storage replica failure & target network delay injection

ksatchit:
-------------------------------------------------

- Adds litmus chaos/test jobs to perform storage replica failure (applicable to containerized storage) & network delay injection on target pods. These jobs are designed to run on pre-deployed applications. 
- Application information (app namespace, label & persistent volume claim name) are the common parameters passed to the test via pod ENV. In addition, there could be test (chaos)-specific parameters (such as chaos specific params such as delay, duration etc.,).
- The chaos library to be used is decided based on the provider storage class (which is derived from the PVC). Currently, the test supports chaos on openebs jiva. 
- The test success criteria depends on the availability of the application through the chaos operation (verified using the application labels) 
- A result custom resource bearing the test name is updated with `phase` & `result` at the start of test & upon completion (`in-progress/none`, `completed/<result, pass|fail>`). 

- The logic to pick the chaoslib based on the storage class is embedded in a jinja template & is executed as a pre-requisite step prior to the actual test
- The litmus result CR is static in nature and updated at specific junctures during the test playbook execution. 
- There is scope for further modularization of the chaoslib, which could be taken up subsequently. 
- A sample test run output of the jobs submitted in this PR are attached for reference: 
   [storage-data-replica-failure.log](https://github.com/openebs/litmus/files/2299296/storage-data-replica-failure.log)
   [storage-target-network-delay.log](https://github.com/openebs/litmus/files/2299544/storage-target-network-delay.log)


apps/percona/chaos/storage_data_replica_failure/chaosutil.j2
```diff
@@ -0,0 +1,5 @@
+{% if sc is defined and sc == 'openebs-standard' %}
```
kmova: name of this file .. can be openebs volume replica failure. -- storage data replica .. might mean percona master/replica.

ksatchit: Done, @kmova

PR #110, opened 2018-08-18
`master.yaml` uses JSONPointer instead of JSONPath

a4abhishek:
**What this PR does**:
This PR makes it easy to marshal `master.yaml` and create patch for litmus tests.

`op`, `path`, `value` are the fields needed to parse JSONPointer, rest of the fields are just for informative purpose.

possible values for `op` (operation) are `"add"`, `"remove"`, `"replace"`, `"move"`, `"test"` and `"copy"`.

**Assumptions:**
- It is assumed in JSONPointer that `ansibletest` container is the first one in YAML when it is parsed.
- Parameters represented in current `master.yaml` is not already present in the litmus test spec. (since I have used `add` in `op` field of each param)

Note: All the above assumptions are met in this PR

Signed-off-by: Abhishek Kashyap <abhishek.kashyap@mayadata.io>

### DIFF
--- a/master.yaml
+++ b/master.yaml
@@ -9,26 +9,42 @@
 tests:
   - name: pod_kill
     test_tags: ['mysql', 'pumba']
-    location: ./apps/percona/tests/mysql_data_persistence/run_litmus_test.yaml
-    test_params:
-      - name: DB_SERVER_IP
-        kind: env
-        type: string
+    specPath: ./apps/percona/tests/mysql_data_persistence/run_litmus_test.yaml
+    patch:
+    - name: 'Server IP'
+      desc: ''
+      kind: env
+      type: string
+      op: 'add'
+      path: '/spec/template/spec/containers/0/env/-'
+      value:
+        name: DB_SERVER_IP
         value: 10.20.10.100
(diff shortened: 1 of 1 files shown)


master.yaml
```diff
@@ -9,26 +9,42 @@
...
-      - name: DB_SERVER_IP
-        kind: env
-        type: string
+    specPath: ./apps/percona/tests/mysql_data_persistence/run_litmus_test.yaml
+    patch:
+    - name: 'Server IP'
+      desc: ''
+      kind: env
+      type: string
+      op: 'add'
```
kmova: should it be add or replace?

a4abhishek: I have checked `apps/percona/tests/mysql_data_persistence/run_litmus_test.yaml` (`specPath`) there is no environment variable named `DB_SERVER_IP` in `ansibletest` container. So, I think it will be `'add'` instead of `'replace'`.

a4abhishek: @ksatchit Then we should update `master.yaml` accordingly in upcoming PRs. Also we can use any of the verbs mentioned in the PR description anytime if we feel so.

PR #115, opened 2018-08-20
Update master.yaml with chaos job info

ksatchit:

- Replace placeholders in master.yaml run template with actual test/chaos job info

### DIFF
--- a/master.yaml
+++ b/master.yaml
@@ -4,10 +4,12 @@
 # Test parameters can be of different kinds such as ENV, configmap etc., with their own properties.
 # The test has tags which will help with categorization & application-mapping.
 
+# This master YAML uses JSONPointer to interpret the params: Refer https://tools.ietf.org/html/rfc6901#section-4
+
 
 tests:
   - name: openebs-volume-replica-failure
-    test_tags: ['mysql', 'openebs']
+    app_tags: ['mysql']
     specPath: ./apps/percona/chaos/openebs_volume_replica_failure/run_litmus_test.yaml
     patch:
     - name: 'Application Namespace'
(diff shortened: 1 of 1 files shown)


master.yaml
```diff
@@ -2,49 +2,80 @@
...
-# will help with categorization & application-mapping
+# The test has tags which will help with categorization & application-mapping.
 
 
 tests:
-  - name: pod_kill
-    test_tags: ['mysql', 'pumba']
-    specPath: ./apps/percona/tests/mysql_data_persistence/run_litmus_test.yaml
+  - name: openebs-volume-replica-failure
+    test_tags: ['mysql', 'openebs']
```
a4abhishek: In this line `openebs` tag refers to what? @ksatchit

master.yaml
```diff
@@ -2,49 +2,80 @@
...
+    specPath: ./apps/percona/chaos/openebs_volume_replica_failure/run_litmus_test.yaml
     patch:
-    - name: 'Server IP'
+    - name: 'APPLICATION NAMESPACE'
```
a4abhishek: I know, there's a reason behind using all uppercase here, but can it happen for any case where mixed case will be better?

Like: Can we consider 'Server IP' over 'SERVER IP' or 'server ip' for any case?

master.yaml
ksatchit: This test uses openebs chaos @a4abhishek . As per my understanding, the tags may reflect key elements of the test metadata (application: mysql, chaostype: openebs). If it were app_tags, then it would have been restricted to mysql.

master.yaml
ksatchit: Can be done.

ksatchit: @a4abhishek , have changed the name descriptors of the tests. Note that the actual ENV names are still maintained under values section.

master.yaml
```diff
@@ -2,49 +2,80 @@
...
+    specPath: ./apps/percona/chaos/openebs_volume_replica_failure/run_litmus_test.yaml
     patch:
-    - name: 'Server IP'
+    - name: 'Application Namespace'
       desc: ''
       kind: env
       type: string
-      op: 'add'
+      op: 'replace'
       path: '/spec/template/spec/containers/0/env/-'
```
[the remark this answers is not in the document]
ksatchit: @a4abhishek , have changed the paths to factor existence of the ENV

master.yaml
```diff
@@ -2,49 +2,80 @@
...
       value:
```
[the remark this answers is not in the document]
ksatchit: Placed just values directly for all params

master.yaml
ksatchit: Changed to app_tags instead of test_tags, with mysql the current app supported

PR #124, opened 2018-08-23
Updates to litmus openebs installer & percona app-deployer jobs

ksatchit:

- The openebs operator installer litmusbook tries to identify node count by performing a `grep <none>` operation with reference to lack of role definitions for the nodes in the cluster. However, this behaviour may change across providers and kubernetes versions. In AWS, for 1.10.x the node roles are explicitly defined causing failure of this step. A more generic approach that checks for taints with 'master' key are used to determine the 'schedulable nodes' of the cluster - thereby getting a more realistic count.

- Simplified a namespace existence check in percona app deployer that was failing due to lack of quotes while performing a string compare within the kubectl jsonpath filter.

### DIFF
--- a/apps/percona/deployers/percona-installer.yaml
+++ b/apps/percona/deployers/percona-installer.yaml
@@ -32,7 +32,8 @@
           when: namespace != 'litmus'
 
         - name: Checking the status  of test specific namespace.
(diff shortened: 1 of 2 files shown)


ashishranjan738:
Nice catch @ksatchit

PR #168, opened 2018-09-19
 Standardize the litmus job & playbook YAML names based on convention 

shashank855:
**This PR does the following**:
- All files names are set to follow common nameing convention.


apps/mongodb/deployers/mongodb_vars.yaml
```diff
@@ -0,0 +1,5 @@
+# Test-specific parametres
+
+mongodb_deployment: mongo-statefulset.yaml
+namespace: litmus
+test_name: Mongodb-Deployment
```
dargasudarshan: @shashank855 Can you change the test name (Invalid value: \"Mongodb-Deployment\": a DNS-1123 subdomain must consist of lower case alphanumeric characters, '-' or '.', and must start and end with an alphanumeric character)

apps/crunchy-postgres/workloads/run_litmus.yml
```diff
@@ -50,4 +50,4 @@ spec:
             value: "200"
 
         command: ["/bin/bash"]
-        args: ["-c", "./test.sh ; exit 0"] 
+        args: ["-c", ".jenkins/deployers/test.sh ; exit 0"]
```
gprasath: Path has to be modified.

gprasath:
@shashank855 As part of this PR, Can we also rename all the files to use delimiter '_' inbetween the words instead of '-'.  cc: @ksatchit

PR #311, opened 2018-10-15
add code to append runid to testname #260

### LINKED ISSUE #260: Update litmus test playbooks to use RUN_ID/instance metadata
## Is this a BUG REPORT or FEATURE REQUEST?
FEATURE REQUEST

**What happened**:

- All litmus jobs update the result of the test in a result custom resource (CR), which derives it's name from the test_name. See below for reference: 

test_name_var: https://github.com/openebs/litmus/blob/master/apps/percona/chaos/openebs_replica_network_delay/test_vars.yml#L1

litmus_cr_template_being_passed_the_test_name: https://github.com/openebs/litmus/blob/master/apps/percona/chaos/openebs_replica_network_delay/test.yml#L118

- The test_name itself has been burned into the ansible playbook so as to reflect, meaningfully, what it intends to do.

- Now, consider the case where the same test is repeated OR executed in parallel, say with different set of params (such as ns, app-labels etc.,) which are provided as ENV configurables . This will cause the re-use of the same custom resource (as the test_name is same) - causing user to consume incorrect result info. 

- This is avoided by providing the litmus jobs with an ability to define "run-instance-metadata" or, say some "run-id" - which can be added as an job/pod ENV - which will be appended to the test_name before creating the litmus result custom resource. Refer: 

```
- block:
      - name: Record test instance/run ID
         set_fact:
            run_id: "{{ lookup('env','RUN_ID') }}"

      - name: Construct testname appended with runID
         set_fact:
            test_name: "{{ test_name }}-{{ run_id }}"

   when: lookup('env','RUN_ID')
```
- Currently, the above logic is available for deployer & chaos litmusbooks, but is not available for other litmusbooks such as workload-generators. 

- All litmusbooks should contain the logic to append RUN_ID - if available to the test_name before deploying the result CR.

**How to fix it (as minimally and precisely as possible)**:

- Audit the litmus test playbooks (typically, `test.yml`) and add the above.

Skchoudhary:
Signed-off-by: Skchoudhary <choudharysaab786@gmail.com>
(cherry picked from commit 2c68fb10e16a6f4b3c291cb5b480926aac185105)

**Which issue this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close that issue when PR gets merged)*: fixes #260 

### DIFF
--- a/apps/jenkins/upgrade/test.yml
+++ b/apps/jenkins/upgrade/test.yml
@@ -7,6 +7,17 @@
 
   tasks:
     - block:
+        - block:
+
+            - name: Record test instance/run ID
+              set_fact:
+                run_id: "{{ lookup('env','RUN_ID') }}"
+
+            - name: Construct testname appended with runID
+              set_fact:
+                test_name: "{{ test_name }}-{{ run_id }}"
+
+          when: lookup('env','RUN_ID')
 
         ## Record the funclib path ##
 


ksatchit:
Thanks for the PR @Skchoudhary !

PR #335, opened 2018-10-26
add openebs callback test

chirag200666:
Fixes issue https://github.com/openebs/litmus/issues/321 

### DIFF
--- a/executor/ansible/plugins/callback/test_openebs.py
+++ b/executor/ansible/plugins/callback/test_openebs.py
@@ -1,0 +1,58 @@
+# Make coding more python3-ish
+import unittest
+from mock import MagicMock
+import json
+from openebs import CallbackModule 
+
+class MagicMock(MagicMock):
+
+    def to_json(self):
+        json = {
+          '_result':{
+              'retries': self._result['retries'],
+              'attempts': self._result['attempts']
+          } 
+        }
+        return json 
+    
+    def __json__(self):
+        return to_json(self)
+
+from json import JSONEncoder
+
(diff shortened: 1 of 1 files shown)


chirag200666:
@ksatchit 
I tried to make the mock work,  but the current solution looks a bit hacky.
I think I'll setup actual callback & then do the test, any better way to do this.

Also what exactly I should asset, since the function doesn't return anything & finally calls another function.

Moteesh:
> @ksatchit
> I tried to make the mock work, but the current solution looks a bit hacky.
> I think I'll setup actual callback & then do the test, any better way to do this.
> 
> Also what exactly I should asset, since the function doesn't return anything & finally calls another function.

@chirag200666 ,
       The mock looks good as there is not much to do in terms of testing the insides of the callback plugin. As we are just printing certain debug messages upon passing the verbosity flag. I see that you are already testing the verbosity by passing the respective flags through the mock object. Using an actual callback is not encouraged in unit testing environment. Use of mock objects is suggested over using actual callbacks.

chirag200666:
@Moteesh 
Thanks for the response. Can you please suggest me, what exactly should I assert or check in the test cases?

Moteesh:
> @Moteesh
> Thanks for the response. Can you please suggest me, what exactly should I assert or check in the test cases?

As there is no return value from the callback, you can get the output and verify it against the know verbose log statement that we want to get from the stdout/stderr of the process. This link might be of help: https://stackoverflow.com/questions/16571150/how-to-capture-stdout-output-from-a-python-function-call

PR #343, opened 2018-10-28
Added Unit test for getEnv function

shivani-1521:
**What this PR does / why we need it**: Unit test for getEnv function

**Which issue this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close that issue when PR gets merged)*: fixes # Issue 301

### DIFF
--- a/pkg/util/envs_test.go
+++ b/pkg/util/envs_test.go
@@ -7,7 +7,7 @@ you may not use this file except in compliance with the License.
 You may obtain a copy of the License at
 
     http://www.apache.org/licenses/LICENSE-2.0
-    
+
 Unless required by applicable law or agreed to in writing, software
 distributed under the License is distributed on an "AS IS" BASIS,
 WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
@@ -24,7 +24,7 @@ import (
 	"testing"
 )
 
-func TestgetEnv(t *Testing.T){
+func TestgetEnv(t *testing.T){
 
 	testCases := []string{
 		"abc",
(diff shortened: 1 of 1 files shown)


pkg/util/envs_test.go
```diff
@@ -0,0 +1,48 @@
...
+	for _, v := range testCases {
+		err := os.Setenv("LITMUS_IO_KUBE_NAMESPACE", v) 
+		if err != nil {
+			t.Fatal("Unable to set environment variable")
+		}
+
+		if(strings(os.Getenv(string("LITMUS_IO_KUBE_NAMESPACE"))) != v){
+			t.Errorf("Expected was %s, got %s", v, strings(os.Getenv(string("LITMUS_IO_KUBE_NAMESPACE"))))
+		}
+	}
```
prateekpandey14: @shivani-1521 this test will not work.. can you do  something like this ..
```go
func TestgetEnv(t *testing.T) {
	testCases := map[string]struct {
		key         string
		value       string
		expectValue string
	}{
		"Missing env variable": {
			key:         "",
			value:       "",
			expectValue: "",
		},
		"Present env variable with value": {
			key:         "LITMUS_IO_KUBE_NAMESPACE",
			value:       "value1",
			expectValue: "value1",
		},
		"Present env variable with empty value": {
			key:         "LITMUS_IO_KUBE_NAMESPACE",
			value:       "",
			expectValue: "",
		},
	}
	for k, v := range testCases {
		t.Run(k, func(t *testing.T) {
			os.Setenv(v.key, v.value)
			actualValue := getEnv(ENVKey(v.key))
			if !reflect.DeepEqual(actualValue, v.expectValue) {
				t.Errorf("expected %s got %s", v.expectValue, actualValue)
			}
			os.Unsetenv(v.key)
		})
	}
}
```

PR #354, opened 2018-10-31
cstor(pool_chaos): check the app status after bringing down pool

vishnuitta:
This PR is to add a litmus job that does the following:
- checks that app is running fine
- verifies for provisioning by looking for cvr and their status
- bring down the cstor pool pod
- checks that app is running fine still
- bring down cstor pool pod again (can be same or different one)
- checks that app is running fine still

This also takes care of ReplicaCount provided in StorageClass.

This doesn't include setting the liveness pod related environment variables, which need to be done along with bash script for e2e.

Signed-off-by: Vishnu Itta <vitta@mayadata.io>


apps/percona/chaos/cstor_pool_failure/run_litmus_test.yml
```diff
@@ -0,0 +1,76 @@
...
+                fieldPath: metadata.name
+          - name: MY_POD_NAMESPACE
+            valueFrom:
+              fieldRef:
+                fieldPath: metadata.namespace
+          # spec.volumes is not supported via downward API
+          - name: MY_POD_HOSTPATH
+            value: /mnt/chaos/cstor_pool_failure
+        command: ["/bin/bash"]
+        args: ["-c", "./logger.sh -d ansibletest -r maya,openebs,pvc,percona; exit 0"]
```
gprasath: @vishnuitta , shall we include 'cstor' string here to catch the logs from those pods too.

chaoslib/openebs/cstor_pool_failure.yaml
```diff
@@ -0,0 +1,73 @@
...
+  register: sc
+
+- name: Derive ReplicaCount from SC
+  shell: >
+    kubectl get sc {{ sc.stdout }} -n {{ operator_ns }} -o json |
+    jq '.metadata.annotations."cas.openebs.io/config"' -r |
+    grep -A 1 -w "name: ReplicaCount" | grep -w value | awk '{print $2}'
+  args:
+    executable: /bin/bash
+  register: replicacount
```
gprasath: Shall we derive this using deployment.

vishnuitta: deployment will not have replicacount

apps/percona/chaos/cstor_pool_failure/run_litmus_test.yml
vishnuitta: yes @gprasath .. it should be good.. will add that

apps/percona/chaos/cstor_pool_failure/test.yml
```diff
@@ -0,0 +1,146 @@
...
+            app: ""
+            phase: completed
+            verdict: "{{ flag }}"
+           
+        - name: Apply the litmus result CR
+          shell: kubectl apply -f litmus-result.yaml
+          args:
+            executable: /bin/bash
+          register: lr_status 
+          failed_when: "lr_status.rc != 0"
```
ksatchit: There is a util we can include to generate and update the result CR now : https://github.com/openebs/litmus/blob/master/common/utils/update_litmus_result_resource.yml

chaoslib/openebs/cstor_pool_failure.yaml
ksatchit: The grep becomes a necessity here as we hit a dead-end upon extracting the content (params) from the cas.openebs.io/config annotation !

chaoslib/openebs/cstor_pool_failure.yaml
```diff
@@ -0,0 +1,73 @@
...
+  register: cvr_count
+
+- name: Compare ReplicaCount and cvr_count to verify provisioning
+  debug:
+    msg:
+      - "replicaCount: {{ replicacount.stdout }}"
+      - "replicaCountInt: {{ replicacount.stdout | int }}"
+      - "replicacnt: {{ replicacnt }}"
+      - "cvr_count: {{ cvr_count.stdout | int }}"
+  failed_when: "replicacnt|int != cvr_count.stdout|int"
```
ksatchit: :+1:

chaoslib/openebs/cstor_pool_failure.yaml
```diff
@@ -0,0 +1,73 @@
+- name: Derive PV from application PVC
```
ksatchit: This is a good util to check volume status/sanity before proceeding to the chaos (pool deployment delete). What do you think about renaming this to cstor_volume_sanity.yaml - we could include this as a pre-chaos step in every cstor test

ksatchit: The cstor pool delete could then be called separately from the test? i.e., the chaosutil selection jinja template could call this instead.

chaoslib/openebs/cstor_pool_failure.yaml
```diff
@@ -0,0 +1,73 @@
...
+
+- name: Check status of cvr
+  command: echo "{{ item }}"
+  failed_when: "item != \"Online\""
+  with_items:
+    - "{{ cvr_status_phase.stdout_lines }}"
+
+- include: cstor_delete_pool_deployment.yaml
+
+- include: cstor_delete_pool_deployment.yaml
```
ksatchit: A good test to do it more than once, as shuf will ensure selection of a different cvr (and thereby pool) most times. Would it be good to define the chaos iteration count in the test_vars & also include a volume sanity check before each iteration of chaos - so we will fail faster, if at all.

tools/ansible-runner/Dockerfile
```diff
@@ -6,7 +6,7 @@ LABEL maintainer="OpenEBS"
 RUN rm -rf /var/lib/apt/lists/* && \
     apt-get clean && \
     apt-get update --fix-missing || true && \
-    apt-get install -y python-minimal python-pip \
+    apt-get install -y jq python-minimal python-pip \
```
ksatchit: We could go ahead with this - just wanted to see if we can use kubectl natively (there are some thoughts around bundling lib/tests so they can downloaded/runnable in/outside containers - i.e., wherever kubeconfig is present)

ksatchit:
Thanks @vishnuitta , this is very helpful! Suggested some minor changes, which can be discussed.

apps/percona/chaos/cstor_pool_failure/test.yml
vishnuitta: done

chaoslib/openebs/cstor_pool_failure.yaml
vishnuitta: I made this as cstor_verify_pool_provisioning.yml, and, included this in cstor_pool_failure.yml.
This file cstor_verify_pool_provisioning.yml can be moved to different place to use this from other directories.

chaoslib/openebs/cstor_pool_failure.yaml
vishnuitta: shuf can give the same cvr also, but, that should be fine for us even if it is different.
This same file verifies that deployment is started again.

tools/ansible-runner/Dockerfile
vishnuitta: removed the dependency on jq

PR #355, opened 2018-11-01
feat(yaml)(liveness): Add external liveness litmus job for percona application

ksatchit:

The liveness jobs need to be enhanced in order to aid launch/monitor from CI tools such as gitlab & made consistent with other "test" job types in litmus (deploy/workload/functional/chaos) 

- Replace a plain K8s job with a litmus job
- Use env vars to accept database cred info (remove conf files, as this necessitates users to run an additional step to create configmap before they can execute the liveness job) 
- Launch liveness container as a sidecar to ansible runner
- Monitor liveness status via ansible runner & update result CRs
- Use placeholders for nodeselector & add taint-toleration to aid infra-chaos tests, where the liveness pod can be isolated by running them on non-participating nodes.

### DIFF
--- a/apps/percona/liveness/test.yml
+++ b/apps/percona/liveness/test.yml
@@ -1,0 +1,64 @@
+---
+- hosts: localhost
+  connection: local
+
+  vars_files:
+    - test_vars.yml
+
+  tasks:
+    - block:
+        - block:
+
+            - name: Record test instance/run ID
+              set_fact:
+                run_id: "{{ lookup('env','RUN_ID') }}"
+
+            - name: Construct testname appended with runID
+              set_fact:
+                test_name: "{{ test_name }}-{{ run_id }}"
+
+          when: lookup('env','RUN_ID')
+        
+        ## RECORD START-OF-TEST IN LITMUS RESULT CR
+        - include_tasks: /common/utils/update_litmus_result_resource.yml
+          vars:
+            status: 'SOT'
+
+        - name: Checking whether liveness container is running
+          shell: >
+            kubectl get pod {{ test_pod }} -n litmus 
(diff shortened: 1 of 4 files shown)


apps/percona/liveness/test.yml
```diff
@@ -0,0 +1,64 @@
...
+            flag: "Fail"
+
+      always:
+
+        ## RECORD END-OF-TEST IN LITMUS RESULT CR
+        - include_tasks: /common/utils/update_litmus_result_resource.yml
+          vars:
+            status: 'EOT'
+
+        ## if flag=fail, task exits immediately; if flag=pass, runs indefinitely until liveness check ends
```
vishnuitta: did we missed the check for 'flag' to be 'pass'?

apps/percona/liveness/test.yml
```diff
@@ -0,0 +1,64 @@
...
+            flag: "Pass"
+
+      rescue:
+        - set_fact:
+            flag: "Fail"
+
+      always:
+
+        ## RECORD END-OF-TEST IN LITMUS RESULT CR
+        - include_tasks: /common/utils/update_litmus_result_resource.yml
```
vishnuitta: should this be at the end?

vishnuitta: is liveness check pod exit treated failure?

ksatchit: Intentionally placed here, at the CR is desired to reflect liveness deploy success/fail.

apps/percona/liveness/test.yml
ksatchit: Thanks @vishnuitta , will add this

apps/percona/liveness/test.yml
vishnuitta: :+1:

PR #380, opened 2018-11-20
(refactor)(yaml): update percona deployment template to include liveness probe

ksatchit:

In Litmus, currently, all stateful applications (i.e., databases) have a corresponding : 

a) "external liveness" check which will verify whether an application is accessible from a client-like end-application

b) "workload/loadgen" jobs which will ensure that application specific data is pumped into the stateful applications. They are being used in litmus/e2e for:

     i) Relative BM measurements
     ii) Fill apps with some data before running functional/chaos tests

However, (b) is short-lived - doesn't run through duration of test. To have active I/Os during chaos, we need "probes" on the application which can keep performing I/O & will accelerate RO conditions. 

- Though named the same (a) is designed in cases where it is not known prior if app has a liveness probe configured or no, i.e., it is a way to ascertain cluster health when a user exposes a health endpoint.

### DIFF
--- a/apps/percona/deployers/percona.yml
+++ b/apps/percona/deployers/percona.yml
@@ -20,7 +20,7 @@ spec:
             limits:
               cpu: 0.5
           name: percona
-          image: percona
+          image: openebs/tests-custom-percona:latest
           imagePullPolicy: IfNotPresent
           args:
             - "--ignore-db-dir"
@@ -34,6 +34,12 @@ spec:
           volumeMounts:
             - mountPath: /var/lib/mysql
               name: data-vol
+          livenessProbe:
+            exec:
+              command: ["bash", "sql-test.sh"]
+            initialDelaySeconds: 60
+            periodSeconds: 1
+            timeoutSeconds: 10
       volumes:
         - name: data-vol
           persistentVolumeClaim:


apps/percona/deployers/percona.yml
```diff
@@ -34,6 +34,12 @@ spec:
           volumeMounts:
             - mountPath: /var/lib/mysql
               name: data-vol
+          livenessProbe:
```
vishnuitta: we may need to set `restartPolicy` as never (or) the checks that we in litmus job for AUT liveness need to have tasks for looking for any restarts.

apps/percona/deployers/percona.yml
```diff
@@ -34,6 +34,12 @@ spec:
...
+            exec:
+              command: ["bash", "sql-test.sh"]
+            initialDelaySeconds: 60
```
vishnuitta: this value should be good. But, thinking whether there is a way to avoid this hard-coded value which can depend on the utilization or H/W specifications of node, something like readiness (or) some checks in sql-test.sh itself to find whether the pod is ready or not.

apps/percona/deployers/percona.yml
ksatchit: The checks done for AUT liveness in litmus chaos jobs considers temporary loss of app endpoint if there are restarts - and has retries with specific intervals. So it will continue to be an indicator of app health. So, in case of spurious restarts by the livenessProbe (due to say, db init is delayed, etc.,) if it can recover and continue to write - it should still be OK. In case of genuine probe failures (such as RO or DB corruption), the restarts will anyways result in app staying in crashLoopBackoff - which is desirable for us -- and in this case the external liveness to will fails as the retry checks will be exhausted

```

File: apps/percona/deployers/percona.yml (hunk 2 above)
Comment (ksatchit) [reply apps/percona/deployers/percona.yml]: Good idea @vishnuitta ! readiness probe is the way to go.  Can I work on this in a separate PR ? 

The sql-test.sh itself has a DB init wait period (which is also hard-coded, btw) before it starts  the periodic mysql `create table-insert data-drop db` routine - which makes it ~660s of max period for DB init. 

```
echo -e "\nWaiting for mysql server to start accepting connections.."
retries=10;wait_retry=60
for i in `seq 1 $retries`; do
  mysql -uroot -pk8sDem0 -e 'status' > /dev/null 2>&1
  rc=$?
  [ $rc -eq 0 ] && break
  sleep $wait_retry
done
```

PR #382, opened 2018-11-20
(refactor)(yaml): update the disk attach playbooks to take disk actions

ksatchit:

- Update the AWS and GPD disk attach playbooks to format/mount conditionally based on a var.
- Disks will be formatted & mounted/unmounted only upon `disk_attach: format` extra_var

### DIFF
--- a/k8s/gcp/gpd-disks/vars.yml
+++ b/k8s/gcp/gpd-disks/vars.yml
@@ -1,0 +1,2 @@
+---
+disk_action: attach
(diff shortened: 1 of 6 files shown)


k8s/gcp/gpd-disks/vars.yml
```diff
@@ -0,0 +1,2 @@
+---
```
vishnuitta: this seems to be new file.. should we include this in other ymls?

ksatchit: EBS disks attach has this already, introduced one for GPD

PR #384, opened 2018-11-21
feat(litmus): Add capacity test cases

utkarshmani1997:
This commit add following test cases:
- Volume creation of 25GB
- Bringing one of the replicas down multiple times randomly
  to create autogenerated snapshots and writing data randomly.
- Ensure the replicas getting connected after and volume is
  in RW mode.

### Note:  
* As per the current build of jiva >= 0.7.0 this test will be failing due to bug in jiva (for random writes).

Signed-off-by: Utkarsh Mani Tripathi <utkarsh.tripathi@mayadata.io>

### DIFF
--- a/apps/fio/tests/capacity/fio.yml
+++ b/apps/fio/tests/capacity/fio.yml
@@ -15,9 +15,10 @@ spec:
         kubernetes.io/hostname: testNode
       containers:
       - name: perfrunner
-        image: openebs/tests-fio:ci
+        image: openebs/tests-fio
         command: ["/bin/bash"]
-        args: ["-c", "./fio_runner.sh --template file/basic-rw --size 256m --duration 60; exit 0"]
+        # value of testduration will be set to zero
+        args: ["-c", "./fio_runner.sh --template file/testtemplate --size testsize --duration testduration; exit 0"]
         volumeMounts:
            - mountPath: /datadir
              name: fio-vol
--- a/apps/fio/tests/capacity/fio.yml
+++ b/apps/fio/tests/capacity/fio.yml
@@ -1,0 +1,29 @@
+---
+apiVersion: batch/v1
+kind: Job
+metadata:
+  name: fio
+spec:
+  template:
+    metadata:
+      name: fio
+      labels:
+        name: fio
+    spec:
+      restartPolicy: Never
+      nodeSelector:
+        kubernetes.io/hostname: testNode
+      containers:
+      - name: perfrunner
+        image: openebs/tests-fio:ci
+        command: ["/bin/bash"]
+        args: ["-c", "./fio_runner.sh --template file/basic-rw --size 256m --duration 60; exit 0"]
+        volumeMounts:
+           - mountPath: /datadir
+             name: fio-vol
+        tty: true
+      volumes:
+      - name: fio-vol
+        persistentVolumeClaim:
+          claimName: testclaim
+---
--- a/apps/fio/tests/capacity/test.yml
+++ b/apps/fio/tests/capacity/test.yml
@@ -30,13 +30,17 @@
 
        - name: Confirm pvc status is bound
          shell: >
-           kubectl get pvc --no-headers -n litmus
+           kubectl get pvc --no-headers -n litmus -l app=test-capacity
          args:
            executable: /bin/bash
          register: result
          until: "'openebs-standard' and 'Bound' in result.stdout"
-         delay: 120
-         retries: 15
+         delay: 30
+         retries: 50
+
+       - name: Wait for controller to get ready
+         wait_for:
+           timeout: 60
 
        ## RUN FIO WORKLOAD TEST
        - name: Populate the iterator with the values to loop
@@ -45,6 +49,7 @@
          args:
            executable: /bin/bash
          register: iterator
+
        - include: test_capacity.yml
          with_items: "{{ iterator.stdout_lines }}"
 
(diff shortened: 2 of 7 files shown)


apps/fio/tests/capacity/test.yml
```diff
@@ -0,0 +1,59 @@
...
+  vars_files:
+    - test_vars.yml
+ 
+  tasks:
+   - block:
+       - name: Replace the pvc placeholder with test param in pvc spec
+         replace:
+           path: "{{ pvc_yaml_alias }}"
+           regexp: "testclaim"
+           replace: "{{ test_name }}"
```
vishnuitta: it seems - we are replacing claim with test_name, which doesn't seem to be correct.

apps/fio/tests/capacity/test.yml
```diff
@@ -0,0 +1,59 @@
...
+         args:
+           executable: /bin/bash
+
+       - name: Confirm pvc status is bound
+         shell: >
+           kubectl get pvc --no-headers -n litmus
+         args:
+           executable: /bin/bash
+         register: result
+         until: "'openebs-standard' and 'Bound' in result.stdout"
```
vishnuitta: this can go wrong result if there is some PV/PVC which are bound and using openebs-standard class. You may need to use -l option

apps/fio/tests/capacity/test.yml
```diff
@@ -0,0 +1,59 @@
...
+         delay: 120
```
vishnuitta: 120 seconds is a huge delay.. lets reduce this to 10 seconds and increase retries accordingly

apps/fio/tests/capacity/test.yml
vishnuitta: ok.. here, you are trying to use test_name itself as claim name.. should be fine

apps/fio/tests/capacity/test_capacity.yml
```diff
@@ -0,0 +1,168 @@
...
+  delay: 120
+  retries: 15
+
+- name: Obtain name of fio pod
+  set_fact:
+    fio_pod_name: "{{ result.stdout.split()[0] }}"
+
+- name: Wait for fio pod to proceed with workload
+  wait_for:
+    timeout: 60
```
vishnuitta: you can make this as 'testduration'

apps/fio/tests/capacity/fio.yml
```diff
@@ -0,0 +1,29 @@
...
+        name: fio
+    spec:
+      restartPolicy: Never
+      nodeSelector:
+        kubernetes.io/hostname: testNode
+      containers:
+      - name: perfrunner
+        image: openebs/tests-fio:ci
+        command: ["/bin/bash"]
+        args: ["-c", "./fio_runner.sh --template file/basic-rw --size 256m --duration 60; exit 0"]
```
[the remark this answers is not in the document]
utkarshmani1997: This is modifiable by environment vars used above this command, and we are modifying it in test_capacity.yaml

apps/fio/tests/capacity/test.yml
utkarshmani1997: roger that

vishnuitta:
capacity used doesn't seem to increase with this as all the iterations of fio will update the same file

apps/fio/tests/capacity/test.yml
utkarshmani1997: sounds good

apps/fio/tests/capacity/test_capacity.yml
```diff
@@ -0,0 +1,164 @@
...
+    | sort | uniq
+  args:
+    executable: /bin/bash
+  register: result
+  failed_when: result.stdout != " 0,"
+
+- name: Get the name of replica pods
+  shell: >
+    kubectl get pods -l openebs.io/replica=jiva-replica
+    -n litmus --no-headers -o name | sed "s/^.\{4\}//"
```
ksatchit: will it help to use jsonpath to derive names

utkarshmani1997: No @ksatchit since we are listing the n no of pods which will be used later to delete replicas randomly. We can use templates for this case also like this 
```kubectl get pods -l openebs.io/replica=jiva-replica                                                                                            
~ 115     -n litmus --template '{{range .items}}{{.metadata.name}}{{"\n"}}{{end}}'```

Comment (ksatchit) [reply apps/fio/tests/capacity/test_capacity.yml]: :+1:

**Comment by ksatchit** [comment]
> capacity used doesn't seem to increase with this as all the iterations of fio will update the same file

@utkarshmani1997 , any comments on above?

**Comment by utkarshmani1997** [comment]
yes @ksatchit if we write randomly in range > 1.5 G we will be able to reproduce the issue. This is enough to generate more than 25k extents. Problem with sequential write is we will not able to generate large no of extents. I think instead of naming it as capacity, we should rename to something else like `test-extents`, what do you suggest ?

File: apps/fio/tests/capacity/test_capacity.yml [hunk 7]
```diff
@@ -0,0 +1,164 @@
...
+    regexp: "testclaim"
+    replace: "{{ test_name }}"
+
+- name: Replace the default I/O test duration with user-defined period
+  replace:
+    path: "{{ pod_yaml_alias }}"
+    regexp: "testduration"
+    replace: "{{ lookup('env','FIO_TESTRUN_PERIOD') }}"
+
+- name: Get controller svc
```
utkarshmani1997: @nsathyaseelan you can make use if this.

PR #394, opened 2018-11-23
network(chaos): adding network ingress network loss on replica 

vishnuitta:
This PR is to add litmus job to induce incoming network packet loss in randomly selected replica.


ksatchit:
Thanks @vishnuitta ! As discussed, we can add checks for volume replica state changes & processing of network chaos type in a future PR.

PR #401, opened 2018-11-26
fix(status): pool chaos status lookup change from Online to Healthy

vishnuitta:
This PR is to have default application namespace as app-percona-ns.
This also fixes the CVR status change from "Online" to "Healthy"

Signed-off-by: Vishnu Itta <vitta@mayadata.io>

### DIFF
--- a/apps/percona/chaos/cstor_pool_failure/run_litmus_test.yml
+++ b/apps/percona/chaos/cstor_pool_failure/run_litmus_test.yml
@@ -32,7 +32,7 @@ spec:
             value: default
 
           - name: APP_NAMESPACE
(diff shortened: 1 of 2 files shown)


apps/percona/chaos/cstor_pool_failure/run_litmus_test.yml
```diff
@@ -32,7 +32,7 @@ spec:
             value: default
 
           - name: APP_NAMESPACE
-            value: percona-cstor
+            value: app-percona-ns
```
ksatchit: assuming this is to default to jiva replica failure

PR #463, opened 2019-01-02
Fix for cstor_pool_scale test

shashank855:
- `kubectl patch` command for scale up/down the pool deployment worked for openshift but failing in pipeline. Instead of patch we can use `kubectl scale`.


apps/percona/chaos/cstor_pool_scale/test.yml
```diff
@@ -85,7 +85,7 @@
           register: cvr_name1  
 
         - name: Scaling-down cstor-pool
-          shell: kubectl patch deployment {{ deployment_name.stdout }} -n {{ operator_ns }} --type=json -p='[{"op":"add", "path":"/spec/replicas", "value":"0"}]'
+          shell: kubectl scale deployment {{ deployment_name.stdout }} -n {{ operator_ns }} --replicas=0 
           register: result1
           failed_when: "'patched' not in result1.stdout"
```
dargasudarshan: @shashank855 The check condition should be changed here as you are scaling 'patched' string wont be there in stdout right ?

apps/percona/chaos/cstor_pool_scale/test.yml
```diff
@@ -117,7 +117,7 @@
           register: cvr_name2   
 
         - name: Scaling-up cstor-pool
-          shell: kubectl patch deployment {{ deployment_name.stdout }} -n {{ operator_ns }} --type=json -p='[{"op":"add", "path":"/spec/replicas", "value":"1"}]'
+          shell: kubectl scale deployment {{ deployment_name.stdout }} -n {{ operator_ns }} --replicas=1
           register: result2
           failed_when: "'patched' not in result2.stdout"
```
dargasudarshan: Same as above comment

apps/percona/chaos/cstor_pool_scale/test.yml
shashank855: @dargasudarshan I made the change.

apps/percona/chaos/cstor_pool_scale/test.yml
shashank855: @dargasudarshan Made the change.

PR #599, opened 2019-04-23
Replaced runAsUser with fsGroup in spec and reduced the volume capacity from 25G to 8G

sushma1118:
* In specification of es-statefulset.yml replaced runAsUser:1000 with fsGroup:1000 .
* Sometimes the pod may get restarted again and again and application pod thus goes to CrashLoopBackOff.
* Replacing solves the issue.
* More volume capacity may sometimes lead to some issues in openShift cluster , so reducing the capacity to 8G may not lead to such issues.

### DIFF
--- a/apps/elasticsearch/deployers/run_litmus_test.yml
+++ b/apps/elasticsearch/deployers/run_litmus_test.yml
@@ -17,7 +17,7 @@ spec:
       containers:
 
       - name: ansibletest
-        image: openebs/ansible-runner:ci
+        image: sushmash/ansible-runner:ci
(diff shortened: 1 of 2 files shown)


apps/elasticsearch/deployers/run_litmus_test.yml
```diff
@@ -17,7 +17,7 @@ spec:
       containers:
 
       - name: ansibletest
-        image: openebs/ansible-runner:ci
+        image: sushmash/ansible-runner:ci
```
gprasath: Can you update the image

PR #665, opened 2019-06-17
Refactor ansiblebook for cluster-cleanup in packet

shashank855:
- Refactor ansiblebook for cluster-cleanup in packet


shashank855:
#666

PR #673, opened 2019-06-26
Changed the amount of data to be filled in disks in order to evict the application pod

sushma1118:
* If availabe size if 50 GB and if we try to fill the disk with more than 50GB, we get and error message 
 ' fallocate: fallocate '/test.img': No space left on device'
* If we fill with 50GB , along with error message, we can observe that the pod is evicted from that node and    gets scheduled on some other node.
* The following file is modified :
   /litmus/experiments/chaos/node_storage_fill/test.yml


ksatchit:
why are we hardcoding values?

sushma1118:
For the time being, it is hardcode, later it will be customized.

Thank you

sushma1118:
For the time being, it is hardcoded. Later it will be customized.

Thank you

PR #705, opened 2019-08-30
refactor[yaml]: Changed the label selector for cspc provision litmusbook

nsathyaseelan:
Signed-off-by: nsathyaseelan <sathyaseelan.n@mayadata.io>

- changed the label selector to get the cspc pool pod status task in litmusbook. As per new label schema changes.


ksatchit:
Let us identify a way to globally define these vars (labels etc.,)

PR #816, opened 2019-10-05
refactor[executor] to patch chaosengine status

shubhamchaudhary:

**-  Executor enhancement to patch chaosengine status**

**Checklist**

* [x] Have you added debug messages where necessary? 
* [x] Have you added task comments where necessary? 
* [x] Have you reused/created util functions instead of repeating tasks in the litmusbook?
* [x] Have you ensured minimum/careful usage of shell utilities (awk, grep, sed, cut, xargs etc.,)?

**limitation**
```
---
kind: ChaosEngine
metadata:
  name: engine-nginx
  namespace: litmus
spec:
  appinfo:
    appns: litmus
    applabel: "run=nginx"
  chaosServiceAccount: litmus

  experiments:
    - name: pod-delete
      spec:
        components:
status:
  experiments:
  - name:   <---   here
 Either We have to provide name or just  "-" (empty list) here so that the value of 
spec.experiments field remains not null otherwise it is unable to add engine status 
using LineInFile or BlockinFile .
```

**Logs of the executor and job**
[logs-job.txt](https://github.com/litmuschaos/litmus/files/3693801/logs-job.txt)
[logs-executor.txt](https://github.com/litmuschaos/litmus/files/3693881/logs-executor.txt)


executor/executor.yml
```diff
@@ -136,4 +136,50 @@
...
+
+- name: patch the chaosengine status
+  lineinfile:
+    dest: chaos-engine.yaml
+    state: present
+    line: '{{item}}'
+  with_items:
+  - "  - name: {{c_experiment_name}}"
+  - "    status: {{c_phase.stdout}}"
+  - "    verdict: {{c_verdict.stdout}}"
```
ksatchit: Can you show an output of how the patched engine looks? The expectation is similar:

```yaml
status:
  experimentStatuses:
     - name:  {{c_experiment_name}}
        instance: {{c_experiment_job_name}} #we can derive this, helps during scheduled chaos
        status: {{c_phase.stdout}}
        verdict: {{c_verdict.stdout}}
```

ksatchit: Also, would it help to have a jinja template with the spec that we can just append using blockinfile?

ksatchit: Also, was this tested with the chaosengine CRD having status as subresource? If this is removed, will the limitation to set an empty experiment list in status be solved?

shubhamchaudhary: > Also, would it help to have a jinja template with the spec that we can just append using blockinfile?

In this approach we have to handle all the dynamic value in jinja template .It will be complex if the env variable are present in engine spec. In that case we have to modify template accordingly

shubhamchaudhary: > Can you show an output of how the patched engine looks? The expectation is similar to below:
> 
> ```yaml
> status:
>   experimentStatuses:
>   - name:  {{c_experiment_name}}
>     instance: {{c_experiment_job_name}} #we can derive this, helps during scheduled chaos
>     status: {{c_phase.stdout}}
>     verdict: {{c_verdict.stdout}}
> ```

Limitation of having empty list can we remove if we use status.experimentStatuses. Initially I was using status.experiments .It has conflict with spec.experiments and blockinfile didn't work as expected due to that conflict

PR #915, opened 2019-11-05
(enhancement) executor: add capability to append an instance_id to chaos result

### LINKED ISSUE #914: Provide ability to inject instance IDs to the chaos result to enable storing details on a per run basis
Currently, the executor generates a unique job hash upon every execution instance of an experiment, but the chaos result continue to be derived purely from static names, i.e., chaosengine & chaos-experiment. There is a need to pass the aforementioned instance ID to the result so as to save state/results for future reference.

This is especially useful in cases of repeated/scheduled executions where the result may differ for the same experiment.

ksatchit:
Signed-off-by: ksatchit <karthik.s@mayadata.io>

- This introduces a way for the executors (in its current form or any custom executors) to propagate an (execution) instance_id of the experiment to its chaos result resource. It could be passed as an ENV, or as an ansible playbook extra var.

**Which issue this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close that issue when PR gets merged)*: fixes #914 

**Checklist**

* [x] Have you provided the positive & negative test logs for the litmusbook execution?
* [x] Have you (jinja) templatized custom scripts that is run by the litmusbook, if any? 
* [x] Have you (jinja) templatized Kubernetes deployment manifests used by the litmusbook, if any?

- The if blocks in the templating logic introduces indentation by default, to avoid which the flag `lstrip_blocks` has been used in the template task. This needs the python package jinja2>2.7. The current ansible-runner uses jinja2 vers 2.10.3.

- The changes at the executor side will be done when needed.

### DIFF
--- a/hack/chaos-result.j2
+++ b/hack/chaos-result.j2
@@ -2,11 +2,14 @@
 apiVersion: litmuschaos.io/v1alpha1
 kind: ChaosResult
 metadata:
-  # name of the litmus testcase
-  name: {{ c_experiment }} 
+  {% if instance_id is defined %}
+  name: {{ c_experiment }}-{{ instance_id }}
+  {% else %}
+  name: {{ c_experiment }}
+  {% endif %} 
 spec:
   # holds the state of testcase,  manually updated by json merge patch
-  # result is the useful value today, but anticipate phase use in future 
+  # verdict is the useful value today, but anticipate phase use in future 
   experimentstatus: 
     phase: {{ phase }}  
     verdict: {{ verdict }} 
--- a/utils/runtime/update_chaos_result_resource.yml
+++ b/utils/runtime/update_chaos_result_resource.yml
@@ -4,6 +4,7 @@
      template:
        src: /chaos-result.j2
        dest: chaos-result.yaml
+       lstrip_blocks: yes
      vars:
        test: "{{ c_experiment }}"
        phase: ""
@@ -23,6 +24,7 @@
      template:
        src: /chaos-result.j2
        dest: chaos-result.yaml
+       lstrip_blocks: yes
      vars:
        test: "{{ c_experiment }}"
        phase: "" 


hack/chaos-result.j2
```diff
@@ -2,11 +2,14 @@
 apiVersion: litmuschaos.io/v1alpha1
 kind: ChaosResult
 metadata:
-  # name of the litmus testcase
-  name: {{ c_experiment }} 
+  {% if instance_id is defined %}
+  name: {{ c_experiment }}{{ instance_id }}
```
shubhamchaudhary: In our naming convention , we append two name by "-" .  Why we don't put  "-" between c_experiment and instance_id.?

ksatchit:
A sample playbook, provided below is used to test these changes, where, the `tmpl.j2` consists of the chaos result template. 

- playbook `test.yml`

```
- hosts: localhost
  connection: local

  vars:
    c_experiment: test
    verdict: pass
    phase: ""

  tasks:
    - name: obtain yaml template
      template:
        src: tmpl.j2
        dest: tmp.yml
        lstrip_blocks: yes
```
- chaosresult template tmpl.j2

```
kind: ChaosResult
metadata:
  {% if instance_id is defined %} 
  name: {{ c_experiment }}{{ instance_id }}
  {% else %}
  name: {{ c_experiment }}
  {% endif %}
spec:
  # holds the state of testcase,  manually updated by json merge patch
  # result is the useful value today, but anticipate phase use in future
  experimentstatus:
    phase: {{ phase }}
    verdict: {{ verdict }}
```

- playbook command executed w/ instance_id defined

```
ansible-playbook test.yml -vv -e "instance_id='-45rc'"
```

- output of the template task

```
kind: ChaosResult
metadata: 

  name: test-45rc
spec:
  # holds the state of testcase,  manually updated by json merge patch
  # result is the useful value today, but anticipate phase use in future
  experimentstatus:
    phase: 
    verdict: pass
```

- ansible playbook command without the instance_id

```
ansible-playbook test.yml -vv
```

- output of the template task

```
kind: ChaosResult
metadata:
  name: test
spec:
  # holds the state of testcase,  manually updated by json merge patch
  # result is the useful value today, but anticipate phase use in future
  experimentstatus:
    phase: 
    verdict: pass
```

hack/chaos-result.j2
ksatchit: Thank you, added!

PR #965, opened 2019-11-26
Remove chaoskube lib from pod-delete experiment

### LINKED ISSUE #963: (chaoslib): remove chaoskube library from pod-kill experiment
@ksatchit told me the chaoskube lib is not supported any more, so I suggest to remove it from the README and the pod-kill experiment. The library I would keep, but give a hint in its files that it is not supported any more.

LaumiH:
**What this PR does / why we need it**: The chaoskube lib is not supported with namespace-scoped SA, so it has to be fully removed from the pod-delete chaos experiment. A notice will be added to the lib under chaoslib/chaoskube.

**Which issue this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close that issue when PR gets merged)*: fixes #963 

**Checklist**

### DIFF
--- a/chaoslib/chaoskube/README.md
+++ b/chaoslib/chaoskube/README.md
@@ -1,3 +1,3 @@
-note: the chaoskube library is not supported by litmuschaos any more. 
-it will be left here, because it can be used if the chaosServiceAccount has cluster-scope.
-as of version 0.8 chaosexperiments are namespace-scoped
+## Note 
+
+The chaoskube library is not supported by litmuschaos any more. It will be left here, because it can be used if the chaosServiceAccount has cluster-scope. As of version 0.8 chaosexperiments are namespace-scoped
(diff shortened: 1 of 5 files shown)


chaoslib/chaoskube/README.md
```diff
@@ -0,0 +1,3 @@
+note: the chaoskube library is not supported by litmuschaos any more. 
+it will be left here, because it can be used if the chaosServiceAccount has cluster-scope.
+as of version 0.8 chaosexperiments are namespace-scoped
```
ksatchit: Can we use some markdown formatting in this README

LaumiH: Of course, I just wanted to give some notice. Do you change it?
Also I noticed that I pushed changes in the experiments/generic/pod_delete/pod_delete_k8s_job.yml I did not want to push, I deleted the default values and commented the service account.

ksatchit: Cool. Looks good. We could just do something like: 

#### Note
The chaoskube library is not supported by litmuschaos any more.  It will be left here, because it can be used if the chaosServiceAccount has cluster-scope. As of version 0.8 chaosexperiments are namespace-scoped

PR #1270, opened 2020-03-07
feat(chaos-result): Adding failStep attribute inside chaosresult

### LINKED ISSUE #1269: Handle events for negative path (experiment failures)
While issue #1244 details approach to generate kubernetes events from the experiments, it deals with the positive experiment workflow. However, events need to handled for the experiment failure scenarios as well. 

A potential approach is detailed here: 

We have accounted only for the positive events here. We need to send out events for the failure cases as well -- whenver they occur (after pre-check/inject/post-check). The approach we are following for this is (thanks for the discussion @Shubham Chaudhary):

- Today, the main logic yaml's always section has only a chaosresult patch with flag (either pass/fail)
The above ^^^ will be enhanced to use a summary.yaml (a generic  util placed inside utils/runtime which is going to: 

  -  look for failures in play till that point (via kubectl logs) and if present
  -  generate a summary event ( i.e., failure, if applicable or just say exp passed)

- The above approach also lends itself to the ease of identifying the following, thereby enabling us to fulfill the pending enhancement on result:
   - failure (step) point in the play
    - recommendation for failure (ansible is already providing this in the failure message)

- This results in an enhancement of the chaosresult status schema (status.experimentStatus.failureStep, `status.experimentStatus.recommendation`) in addition to phase/verdict (for pass case, these will be 'n/a') 

- The chaos-runner will be making the above patch into ChaosEngineStatus with additional fields.

ispeakc0de:

- Adding failStep attribute inside chaosresult (.status.experimentStatuses) which will tell the task-name on which playbook has failed.

- Moving chaos-result.j2 from hacks/ to utils/runtime , as runtime contains all the chaos-result related files

- chaos-result manifest after the changes

```yaml
kind: ChaosResult
metadata:
  name: engine-pod-delete
  labels:
    chaosUID: 83c439de-6082-11ea-89a5-0cc47ab587b8
spec:
  engine: engien
  experiment: pod-delete
status:
  experimentstatus: 
    phase: Completed
    verdict: Pass
    failStep: N/A
   ```
**Which issue this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close that issue when PR gets merged)*: fixes #1269 

**Checklist**

### DIFF
--- a/experiments/coredns/pod_delete/pod_delete_ansible_logic.yml
+++ b/experiments/coredns/pod_delete/pod_delete_ansible_logic.yml
@@ -15,6 +15,7 @@
     kill_count: "{{ lookup('env','KILL_COUNT') }}"
     chaos_uid: "{{ lookup('env','CHAOS_UID') }}"
     c_engine: "{{ lookup('env','CHAOSENGINE') }}"
+    chaos_pod_name: "{{ lookup('env','POD_NAME') }}"
 
   tasks:
     - block:
@@ -129,7 +130,10 @@
           when: "liveness_app_deploy is defined and liveness_app_deploy != '' "
 
       always:
- 
+
+        ## Getting failure step from experiment-pod
+        - include_tasks: /utils/runtime/getting_failure_step.yml  
+        
         ## RECORD END-OF-TEST IN LITMUSCHAOS RESULT CR
         - include_tasks: /utils/runtime/update_chaos_result_resource.yml
           vars:
--- a/utils/runtime/getting_failure_step.yml
+++ b/utils/runtime/getting_failure_step.yml
@@ -1,0 +1,27 @@
+- block:
+
+    ## Getting failure step from experiment-pod
+    - name: Getting failure step from experiment pod
+      shell: >
+        kubectl logs {{ chaos_pod_name }} -n {{ a_ns }} | grep "FAILED!" -B 1 | head -1 | awk -F "Step:" '{print $2}'
+      register: fail_step
+      args: 
+        executable: /bin/bash
+
+    - name: Record the name of failed step
+      set_fact:
+        failStep: "{{ fail_step.stdout }}"
+
+    - name: Print the name of failed step
+      debug:
+        msg: "failStep: {{ failStep }}"
+
+  when: "flag is defined and flag == 'Fail'"
+
+- block: 
+  
+    - name: Record the name of failed step
+      set_fact:
+        failStep: "N/A"
+
+  when: "flag is defined and flag == 'Pass'"
--- a/experiments/generic/container_kill/container_kill_ansible_logic.yml
+++ b/experiments/generic/container_kill/container_kill_ansible_logic.yml
@@ -12,6 +12,7 @@
     lib_image: "{{ lookup('env','LIB_IMAGE') }}"
     chaos_uid: "{{ lookup('env','CHAOS_UID') }}"
     c_engine: "{{ lookup('env','CHAOSENGINE') }}"
+    chaos_pod_name: "{{ lookup('env','POD_NAME') }}"
 
   tasks:
     - block:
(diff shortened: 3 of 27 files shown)

### CODE REVIEW

File: experiments/coredns/pod_delete/pod_delete_ansible_logic.yml [hunk 1]
```diff
@@ -129,6 +130,34 @@
           when: "liveness_app_deploy is defined and liveness_app_deploy != '' "
 
       always:
+
+        
+        - block:
+
+            ## Getting failure step from experiment-pod
+            - name: Getting failure step from experiment pod
+              shell: kubectl logs {{ chaos_pod_name }} -n {{ a_ns }} | grep "FAILED!" -B 1 | head -1 | awk -F "Step:" '{print $2}'
```
ksatchit: Can we use a folded scalar here `>`

experiments/coredns/pod_delete/pod_delete_ansible_logic.yml
```diff
@@ -129,6 +130,34 @@
...
+              register: fail_step
+              args: 
+                executable: /bin/bash
+
+            - name: Record the fail-step and recommandation
```
ksatchit: No recommendation

experiments/coredns/pod_delete/pod_delete_ansible_logic.yml
```diff
@@ -129,6 +130,34 @@
...
+              set_fact:
+                failStep: "{{ fail_step.stdout }}"
+  
+            - name: Print values
```
ksatchit: Can we have an improved task banner

utils/runtime/getting_failure_step.yml
```diff
@@ -0,0 +1,27 @@
...
+
+  when: "flag is defined and flag == 'Fail'"
+
+- block: 
+  
+    - name: Record the name of failed step
+      set_fact:
+        failStep: "N/A"
+
+  when: "flag is defined and flag == 'Pass'"
```
uditgaurav: If it is a single task can we write when with the task itself?

utils/runtime/getting_failure_step.yml
```diff
@@ -0,0 +1,27 @@
+- block:
+
+    ## Getting failure step from experiment-pod
+    - name: Getting failure step from experiment pod
+      shell: >
+        kubectl logs {{ chaos_pod_name }} -n {{ a_ns }} | grep "FAILED!" -B 1 | head -1 | awk -F "Step:" '{print $2}'
+      register: fail_step
+      args: 
+        executable: /bin/bash
```
uditgaurav: Not compulsory, but can we use the same format(order) as we use for all the `shell` commands
```yaml
    ## Getting failure step from experiment-pod
    - name: Getting failure step from experiment pod
      shell: >
        kubectl logs {{ chaos_pod_name }} -n {{ a_ns }} | grep "FAILED!" -B 1 | head -1 | awk -F "Step:" '{print $2}'
      args: 
        executable: /bin/bash
      register: fail_step
```

utils/runtime/getting_failure_step.yml
```diff
@@ -0,0 +1,27 @@
+- block:
+
+    ## Getting failure step from experiment-pod
+    - name: Getting failure step from experiment pod
+      shell: >
+        kubectl logs {{ chaos_pod_name }} -n {{ a_ns }} | grep "FAILED!" -B 1 | head -1 | awk -F "Step:" '{print $2}'
```
uditgaurav: why we don't need `-f` here `kubectl logs -f`?

utils/runtime/getting_failure_step.yml
ksatchit: There will be upcoming tasks added into that block(events)

utils/runtime/getting_failure_step.yml
ksatchit: -f is for follow/tail log.. Since we are already in execution flow in this job, we it need to get logs until that point w/ a command that completes.

PR #1286, opened 2020-03-10
update(chaos-result) Adding condition for c_result name if experiment run via job

ispeakc0de:

- If the c_result is not defined then name of the chaos-result will be same as c_experiment. It is the possible case where we ran experiment via job

**Checklist**

### DIFF
--- a/utils/runtime/chaos-result.j2
+++ b/utils/runtime/chaos-result.j2
@@ -2,13 +2,18 @@
 apiVersion: litmuschaos.io/v1alpha1
 kind: ChaosResult
 metadata:
-  {% if instance_id is defined %}
+  {% if instance_id is defined and c_result is defined %}
   name: {{ c_result }}-{{ instance_id }}
-  {% else %}
+  {% elif c_result is defined %}
   name: {{ c_result }}
+  {% else %}
+  name: {{ c_experiment }}
   {% endif %} 
   labels:
+    {% if chaos_uid is defined %}
     chaosUID: {{ chaos_uid }}
+    {% endif %} 
+    type: ChaosResult
 spec:
   engine: {{ c_engine }}
   experiment: {{ c_experiment }}


utils/runtime/chaos-result.j2
```diff
@@ -2,10 +2,12 @@
...
+  {% if instance_id is defined and c_result is defined %}
   name: {{ c_result }}-{{ instance_id }}
-  {% else %}
+  {% elif c_result is defined %}
   name: {{ c_result }}
+  {% else %}
+  name: {{ c_experiment }}
   {% endif %} 
   labels:
     chaosUID: {{ chaos_uid }}
```
rajdas98: Please also add a condition to check chaos_uid. if engine is not present.

rajdas98:
Closing the PR https://github.com/litmuschaos/litmus/pull/1287. As this PR is similar

utils/runtime/chaos-result.j2
```diff
@@ -2,13 +2,17 @@
...
-  {% else %}
+  {% elif c_result is defined %}
   name: {{ c_result }}
+  {% else %}
+  name: {{ c_experiment }}
   {% endif %} 
   labels:
+    {% if chaos_uid is defined %}
     chaosUID: {{ chaos_uid }}
+    {% endif %}
```
rajdas98: Please put the labels: after the if conditions for now. Otherwise it will failed to create the chaosresult. In future if new labels will add this can be outside of the if conditions

ispeakc0de: Added `name=chaosResult` label for this. It is good to have a label

rajdas98: Otherwise you can generate and random string for chaosUID: {{ chaos_uid }}

ispeakc0de: If we run an experiment via job then there is no point of creating chaos_uid label. It is only needed for the cleanup purpose when we run exp via chaos_engine.

utils/runtime/chaos-result.j2
```diff
@@ -2,13 +2,18 @@
...
+  {% elif c_result is defined %}
   name: {{ c_result }}
+  {% else %}
+  name: {{ c_experiment }}
   {% endif %} 
   labels:
+    {% if chaos_uid is defined %}
     chaosUID: {{ chaos_uid }}
+    {% endif %} 
+    name: ChaosResult
```
ksatchit: type/artifact: ChaosResult

PR #1290, opened 2020-03-10
refactor(exp-job) Removing chaosengine env from experiment-job

ispeakc0de:

- Removing `CHAOS_ENGINE` var from the ansible_job.yml of all experiments as it not useful if we ran experiment via job.

**Checklist**


experiments/coredns/pod_delete/pod_delete_k8s_job.yml
```diff
@@ -53,9 +53,6 @@ spec:
           - name: LIB
             value: ''
 
-          - name: CHAOSENGINE
```
chandankumar4: will it not affect `c_engine: "{{ lookup('env','CHAOSENGINE') }}"` in file experiments/coredns/pod_delete/pod_delete_ansible_logic.yml

ispeakc0de: No, it will not affect it as we already handled this situation in our experiments

chandankumar4: okay!! Then do we need that variable if we are not passing any `CHAOSENGINE` env?

chandankumar4: I think we are using this env to construct the result name, If this env is not passed then how the result name is generated

ispeakc0de: We are using `CHAOSENGINE` env for the generation of logs so it is required only if we run it via chaos-runner but we don't pass it. It will not generate events

ksatchit: When chaos-runner creates this experiment job, it injects the CHAOSENGINE env. Execution as a pure manual job doesn't need the env. The chaosresult jinja template has been equipped to handle these conditions.

PR #1698, opened 2020-07-09
(feat auth): Redirection and Namespace

Sanjay1611:

This PR does following -

- removes the redirection that was being used for the development purposes

- changes namespace to `litmus`

**Checklist**

### DIFF
--- a/litmus-portal/backend/auth/pkg/types/const.go
+++ b/litmus-portal/backend/auth/pkg/types/const.go
@@ -5,7 +5,7 @@ import "time"
 // define the type of authorization request
 const (
 	RedirectURI                string = "http://localhost:3000/"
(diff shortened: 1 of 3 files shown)


litmus-portal/backend/auth/pkg/types/const.go
```diff
@@ -5,7 +5,7 @@ import "time"
 // define the type of authorization request
 const (
 	RedirectURI                string = "http://localhost:3000/"
-	DefaultDBServerURL         string = "mongodb://mongo-service.default.svc.cluster.local:27017"
+	DefaultDBServerURL         string = "mongodb://mongo-service.litmus.svc.cluster.local:27017"
```
gdsoumya, line 8: If all the components are deployed in the same namespace, I think we can just use the name of the service and not the full url. `eg. "mongodb://mongo-service:27017"` or something similar to this. Can you try this.

PR #1759, opened 2020-07-20
chore(litmus-portal): Community Page and Time Series Plot Refractor.

ishangupta-ds:
Signed-off-by: ishangupta-ds <ishan.gupta@mayadata.io>

**What this PR does / why we need it**:

Code refactoring of Community Page and Time Series Plot.

**Which issue this PR fixes**:

fixes #1785 

**Checklist**

* [x] Does this PR have a corresponding GitHub issue?
* [ ] Have you included relevant README for the chaoslib/experiment with details?
* [x] Have you added debug messages where necessary? 
* [x] Have you added task comments where necessary? 
* [x] Have you tested the changes for possible failure conditions?
* [ ] Have you provided the positive & negative test logs for the litmusbook execution?
* [ ] Does the litmusbook ensure idempotency of cluster state?, i.e., is cluster restored to original state?
* [ ] Have you used non-shell/command modules for Kubernetes tasks?
* [ ] Have you (jinja) templatized custom scripts that is run by the litmusbook, if any? 
* [ ] Have you (jinja) templatized Kubernetes deployment manifests used by the litmusbook, if any?
* [ ] Have you reused/created util functions instead of repeating tasks in the litmusbook?
* [x] Do the artifacts follow the appropriate directory structure? 
* [ ] Have you isolated storage (eg: OpenEBS) specific implementations, checks? 
* [ ] Have you isolated platform (eg: baremetal kubeadm/openshift/aws/gcloud) specific implementations, checks?  
* [ ] Are the ansible facts well defined? Is the scope explicitly set for playbook & included utils?
* [ ] Have you ensured minimum/careful usage of shell utilities (awk, grep, sed, cut, xargs etc.,)?
* [ ] Can the limtusbook be executed both from within & outside a container (configurable paths, no hardcode)?
* [ ] Can you suggest the minimal resource requirements for the litmusbook execution?
* [ ] Does the litmusbook job artifact carry comments/default options/range for the ENV tunables?
* [ ] Has the litmusbooks been linted? 

**Special notes for your reviewer**:

PR #1760, opened 2020-07-20
chore(litmus-portal): Header and BreadCrumbs Refractor.

### LINKED ISSUE #1785: Refractor Header and sub components. Refractor time-series graph for community page for Litmus-Portal.
Refractor Header and sub components. Refractor time-series graph for community page for Litmus-Portal.

ishangupta-ds:

Code refactoring of header and BreadCrumbs.

**Which issue this PR fixes**:

fixes #1785 

**Checklist**

* [x] Have you added debug messages where necessary? 
* [x] Have you added task comments where necessary? 
* [x] Do the artifacts follow the appropriate directory structure? 


ishangupta-ds:
@arkajyotiMukherjee , @gdsoumya and @rajdas98 can you review this one after PR #1759

litmus-portal/frontend/src/components/Header/NotificationDropdown.tsx
```diff
@@ -78,6 +77,7 @@ function NotificationsPopperButton(props: NotifierProps) {
           horizontal: 'right',
         }}
         classes={{ paper: classes.popoverPaper }}
+        style={{ marginTop: 17 }}
```
arkajyotiMukherjee, line 80: values in px....use rem

ishangupta-ds:
> PR contains older commits
> Time series plots files are included in the PR

okay i am removing that commit.

ishangupta-ds:
> PR contains older commits
> Time series plots files are included in the PR

That commit has been reverted now. @arkajyotiMukherjee

litmus-portal/frontend/src/components/Header/styles.ts
```diff
@@ -6,13 +6,173 @@ const useStyles = makeStyles((theme: Theme) => ({
...
+    justifyContent: 'space-between',
+    backgroundColor: theme.palette.background.default,
+    marginTop: theme.spacing(0.5),
+    marginBottom: theme.spacing(0),
+    paddingTop: theme.spacing(2),
+    paddingBottom: theme.spacing(0),
   },
-  notificationSection: {
+
+  avatarBackground: {
```
arkajyotiMukherjee, line 56: Please add comments for which section the styles belong to

ishangupta-ds: sure.

litmus-portal/frontend/src/components/Header/NotificationDropdown.tsx
ishangupta-ds: changing to rem.

ankushChatterjee:
I was browsing through the repo and found this PR. This seems like a good project.
@ishangupta-ds If you don't mind me asking I was going through this PR and found that you used useCallback in a lot of places that only contained a single item in the dependency array, would you please care to explain your thought the process for this use of useCallback everywhere.

A good example would be `NotificationItemList.tsx Line No: 27`

Thanks.

arkajyotiMukherjee:
@ankushChatterjee thanks for taking interest in your project. The observation you made makes sense, I too couldn't understand the purpose of that use. @ishangupta-ds will get back to you at his earliest.

ishangupta-ds:
> I was browsing through the repo and found this PR. This seems like a good project.
> @ishangupta-ds If you don't mind me asking I was going through this PR and found that you used useCallback in a lot of places that only contained a single item in the dependency array, would you please care to explain your thought the process for this use of useCallback everywhere.
> 
> A good example would be `NotificationItemList.tsx Line No: 27`
> 
> Thanks.

Yeah, i have made that change purposely with this PR, it has been done after a refactoring requested by the backend team because of the problems faced during integration, now they can simply access the required states of all subcomponents from the main component file itself i.e. index.tsx. cc: @gdsoumya

arkajyotiMukherjee:
@ishangupta-ds I think what @ankushChatterjee is mentioning is the reasoning behind using useCallback() hook in the frontend. I think this is unrelated to backend or code refactoring.
Like I noticed it after he mentioned, in NotificationListItem.tsc in line 27 you have used useCallback() with one dependency into its deps array. Would you care to explain the reasoning behind this choice in the code and why not a simple function.

ishangupta-ds:
> @ishangupta-ds I think what @ankushChatterjee is mentioning is the reasoning behind using useCallback() hook in the frontend. I think this is unrelated to backend or code refactoring.
> Like I noticed it after he mentioned, in NotificationListItem.tsc in line 27 you have used useCallback() with one dependency into its deps array. Would you care to explain the reasoning behind this choice in the code and why not a simple function.

I have recursively passed the function-prop to the useCallback and specified the dependency, then the useCallback hook returns a memoized version of the function-prop that has been passed to the parent component. @arkajyotiMukherjee @ankushChatterjee 

For this particular callback you have pointed out, its only use is to handle error in image loading. Refer to other callback usages for the context.

arkajyotiMukherjee:
@ishangupta-ds 
the definition of useCallback goes as follows:

> Pass an inline callback and an array of dependencies. useCallback will return a memoized version of the callback that only changes if one of the dependencies has changed. This is useful when passing callbacks to optimized child components that rely on reference equality to prevent unnecessary renders (e.g. shouldComponentUpdate).
> 
> useCallback(fn, deps) is equivalent to useMemo(() => fn, deps).

This means that the callback will only fire when one of these dependencies has been changed. using it for setState purpose doesn't quick makes sense as a memorized version of a dispatch method is redundant. Please check the docs and change it to a normal function or an arrow function.

In simple words "If it's meant to fire every time let it fire every time". We would not want a situation where a setState doesn't fire. and React automatically checks under the hood if the value of the previous and new state is the same or not and renders components only if the value changes.

ishangupta-ds:
> @ishangupta-ds
> the definition of useCallback goes as follows:
> 
> > Pass an inline callback and an array of dependencies. useCallback will return a memoized version of the callback that only changes if one of the dependencies has changed. This is useful when passing callbacks to optimized child components that rely on reference equality to prevent unnecessary renders (e.g. shouldComponentUpdate).
> > useCallback(fn, deps) is equivalent to useMemo(() => fn, deps).
> 
> This means that the callback will only fire when one of these dependencies has been changed. using it for setState purpose doesn't quick makes sense as a memorized version of a dispatch method is redundant. Please check the docs and change it to a normal function or an arrow function.
> 
> In simple words "If it's meant to fire every time let it fire every time". We would not want a situation where a setState doesn't fire. and React automatically checks under the hood if the value of the previous and new state is the same or not and renders components only if the value changes.

Yeah the CallBack is redundant here, it can be replaced by a normal function too. React re-renders on state change after comparing the states internally..acknowledged. I will be making the necessary changes.

ishangupta-ds:
> remove unnecessary useCallbacks

I have made changes recently can you review again?

ishangupta-ds:
Thanks ! @arkajyotiMukherjee and @ankushChatterjee for the review. Changes have been made for removing the redundant Callbacks.

litmus-portal/frontend/src/components/Header/NotificationListItem.tsx
```diff
@@ -14,23 +14,19 @@ import DeleteOutlineTwoToneIcon from '@material-ui/icons/DeleteOutlineTwoTone';
 interface NotificationListItemProps {
   message: any;
   divider: boolean;
-  total: number;
+  CallbackOnDeleteNotification: any;
```
arkajyotiMukherjee, line 17: don't use any here

litmus-portal/frontend/src/components/Header/NotificationListItem.tsx
```diff
@@ -14,23 +14,19 @@ import DeleteOutlineTwoToneIcon from '@material-ui/icons/DeleteOutlineTwoTone';
 interface NotificationListItemProps {
   message: any;
```
arkajyotiMukherjee, line 15: don't use any

litmus-portal/frontend/src/components/Header/ProfileDropdownItems.tsx
```diff
@@ -1,54 +1,60 @@
...
-    root: {
-      width: 24,
-      height: 24,
-      border: `3px solid ${theme.palette.background.paper}`,
-    },
-  })
-)(Avatar);
-
-interface ProfileInfoDropdownProps {
+interface ProfileInfoDropdownItemProps {
```
arkajyotiMukherjee, line 19: this whole interface has any type

litmus-portal/frontend/src/components/Header/NotificationListItem.tsx
ishangupta-ds: Changed.

litmus-portal/frontend/src/components/Header/NotificationListItem.tsx
ishangupta-ds: Changed.

litmus-portal/frontend/src/components/Header/ProfileDropdownItems.tsx
ishangupta-ds: Modified.

ishangupta-ds:
All suggested changes have been made. @arkajyotiMukherjee and @S-ayanide

ishangupta-ds:
The types for header have been placed in models. @arkajyotiMukherjee

ishangupta-ds:
@rajdas98 PTAL.

ishangupta-ds:
@arkajyotiMukherjee @gdsoumya @rajdas98 PTAL.

litmus-portal/frontend/src/components/Header/NotificationDropdown.tsx
```diff
@@ -111,8 +113,8 @@ function NotificationsPopperButton(props: NotifierProps) {
   );
 }
 
-NotificationsPopperButton.propTypes = {
+NotificationsDropdown.propTypes = {
```
S-ayanide, line 116: Do you need to write it like this? You already have a NotificationsDropwdownProps Interface at the top, and message param is already present there as required. You can just modify the type to Array of Object right there

litmus-portal/frontend/src/components/Header/NotificationDropdown.tsx
```diff
@@ -14,33 +14,39 @@ import {
...
-  messages: any;
-  count: any;
+interface NotificationDropdownProps {
+  messages: Message[];
+  count: string;
+  CallbackToHeaderOnDeleteNotification: NotificationsCallBackType;
 }
 
-function NotificationsPopperButton(props: NotifierProps) {
+function NotificationsDropdown(props: NotificationDropdownProps) {
```
S-ayanide, line 27: Since we are following the standards of writing a functional component with React.FC, you can rewrite this as
```
const NotificationsDropdown:React.FC<NotificationDropdownProps> = ({ 
messages, 
count, 
CallbackToHeaderOnDeleteNotification }) => {
```

litmus-portal/frontend/src/components/Header/styles.ts
```diff
@@ -1,18 +1,184 @@
...
     width: '100%',
     height: '100%',
     backgroundColor: theme.palette.background.default,
   },
-  breadCrumbsGrid: {
-    marginTop: theme.spacing(2),
-    paddingLeft: theme.spacing(4),
+
+  headerFlex: {
+    height: '5%',
```
Saranya-jena, line 12: I guess we can't use '%' if it's not '100%' according to the guidelines.

litmus-portal/frontend/src/components/Header/NotificationDropdown.tsx
S-ayanide: This is something you did in JavaScript actually, with TS you have more control over your types so although this might work this isn't ideally the right practice I think.

litmus-portal/frontend/src/components/Header/styles.ts
```diff
@@ -1,18 +1,184 @@
...
+    padding: theme.spacing(2),
+    flexDirection: 'column',
+    alignItems: 'center',
+    overflowY: 'auto',
+    maxHeight: theme.spacing(70),
+  },
+
+  tabContainerProfileDropdownItem: {
+    overflowY: 'auto',
+    maxHeight: 350,
```
Saranya-jena, line 87: use rem

litmus-portal/frontend/src/components/Header/styles.ts
```diff
@@ -1,18 +1,184 @@
...
+  userInfo: {
+    display: 'flex',
+    flexDirection: 'column',
+    alignItems: 'center',
+    paddingLeft: theme.spacing(1),
+  },
+
+  userName: {
+    fontSize: 14,
+  },
```
Saranya-jena, line 103: same here..

litmus-portal/frontend/src/components/Header/styles.ts
```diff
@@ -1,18 +1,184 @@
...
+
+  userEmail: {
+    fontSize: 12,
+    color: 'grey',
```
Saranya-jena, line 107: use rem for fontSize

litmus-portal/frontend/src/components/Header/ProfileDropdownItems.tsx
```diff
@@ -1,54 +1,65 @@
...
-
-interface ProfileInfoDropdownProps {
-  anchorEl: any;
-  isOpen: any;
-  onClose: any;
-  name: any;
-  email: any;
-  projects: any;
+interface OnCloseType {
+  (): void;
```
S-ayanide, line 21: OnCloseType only has one param of an anonymous function. If you want to keep it that way, its fine but give it some name, like `func: () => void ;` but since you only have one and you are calling it on the `ProfileInfoDropdownItemProps` interface, it makes more sense to just add `onClose: () => void;` right there

litmus-portal/frontend/src/components/Header/styles.ts
```diff
@@ -1,18 +1,184 @@
...
+    paddingTop: theme.spacing(1),
+    paddingBottom: theme.spacing(1),
+    paddingLeft: theme.spacing(9),
+    paddingRight: theme.spacing(9.5),
+  },
+
+  buttonEditProfile: {
+    borderRadius: theme.spacing(0.3),
+    padding: theme.spacing(0.5, 2),
+    fontSize: 12,
```
Saranya-jena, line 133: here also

litmus-portal/frontend/src/components/Header/styles.ts
```diff
@@ -1,18 +1,184 @@
...
+    marginBottom: 0,
+    paddingTop: 0,
+    paddingBottom: 0,
+  },
+
+  // Styles for NotificationDropdown.
+  tabContainer: {
+    overflowY: 'auto',
+    maxHeight: 350,
+  },
```
Saranya-jena, line 162: here

litmus-portal/frontend/src/components/Header/styles.ts
```diff
@@ -1,18 +1,184 @@
...
+
+  popoverPaper: {
+    width: '100%',
+    maxWidth: 350,
+    marginTop: '1.0625rem',
+    marginLeft: theme.spacing(2),
+    marginRight: theme.spacing(1),
+    [theme.breakpoints.down('sm')]: {
+      maxWidth: 270,
```
Saranya-jena, line 171: here

litmus-portal/frontend/src/components/Header/styles.ts
```diff
@@ -1,18 +1,184 @@
...
+    },
+  },
+
+  divider: {
+    marginTop: -2,
```
Saranya-jena, line 176: use theme.spacing

litmus-portal/frontend/src/components/Header/ProfileDropdownSection.tsx
```diff
@@ -0,0 +1,114 @@
...
+interface ProfileInfoDropdownSectionProps {
+  name: string;
+  email: string;
+  username: string;
+  projects: Project[];
+  selectedProjectID: string;
+  CallbackToSetSelectedProjectID: ProjectsCallBackType;
+}
+
+const ProfileDropdownSection = (props: ProfileInfoDropdownSectionProps) => {
```
S-ayanide, line 20: It'd be better if you could follow the TS standards like React.FC<> here as well

litmus-portal/frontend/src/components/Header/ProjectListItem.tsx
```diff
@@ -11,33 +11,32 @@ import InsertDriveFileOutlinedIcon from '@material-ui/icons/InsertDriveFileOutli
...
 interface ProjectListItemProps {
-  project: any;
-  divider: any;
+  project: Project;
+  divider: boolean;
+  selectedProjectID: string;
+  callbackToSetActiveProjectID: ProjectsCallBackType;
 }
 
 function ProjectListItem(props: ProjectListItemProps) {
```
S-ayanide, line 23: React.FC here as well

litmus-portal/frontend/src/components/Header/index.tsx
```diff
@@ -1,75 +1,106 @@
...
 import CustomBreadCrumbs from '../CustomBreadCrumbs';
-import NotificationsPopperButton from '../NotificationDropdown';
-import ProfileDropdown from '../ProfileDropdown';
+import NotificationsDropdown from './NotificationDropdown';
+import ProfileDropdownSection from './ProfileDropdownSection';
+import { UserData } from '../../models/user';
+import { RootState } from '../../redux/reducers';
+import { NotificationIds, Project, Message } from '../../models/header';
 
 const Header = () => {
```
S-ayanide, line 15: If you have a functional component with does not have a prop taken from any interface, you can write 
```
const Header:React.FC = () => {}
```

litmus-portal/frontend/src/components/Header/styles.ts
S-ayanide: I don't think that's the case. Actually to make it flexible with the screen resolutions it's better to use %. theme.spacing wouldn't be very gradual.

ishangupta-ds:
All the suggested changes have been made. @Saranya-jena and @S-ayanide PTAL.

ishangupta-ds:
@arkajyotiMukherjee @gdsoumya @rajdas98 @amityt @Jonsy13 @oumkale PTAL.

PR #1772, opened 2020-07-22
chore(litmus-portal): Adding dynamic and discovery client pkg to graphql-server

rajdas98:
- Adding discovery client and dynamic client in graphql-server
- Other microservices can use this pkg to generate dynamic and discovery client


litmus-portal/backend/graphql-server/pkg/k8s/client/dynamic/incluster.go
```diff
@@ -0,0 +1,31 @@
+package dynamic
```
gdsoumya, line 1: Do we need this in the GQL Server?

rajdas98: Yes, this pkg will be used by the other microservices like agent/selfdeployer/graphqlserver

PR #1771, opened 2020-07-22
chore(litmus-portal): Updated GQL Sever

gdsoumya:
Signed-off-by: Soumya Ghosh Dastidar <gdsoumya@gmail.com>

**What this PR does / why we need it**:

Features Implemented : 
- Basic MongoDB setup
- Custer Schema defined
- Cluster Registration feature
- Cluster Live/Active Check and Notification
- Updated and optimised Docker image

**Which issue this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close that issue when PR gets merged)*: fixes #

fixes #1768, fixes #1676 

**Checklist**

* [x] Does this PR have a corresponding GitHub issue?
* [ ] Have you included relevant README for the chaoslib/experiment with details?
* [ ] Have you added debug messages where necessary? 
* [ ] Have you added task comments where necessary? 
* [x] Have you tested the changes for possible failure conditions?
* [ ] Have you provided the positive & negative test logs for the litmusbook execution?
* [ ] Does the litmusbook ensure idempotency of cluster state?, i.e., is cluster restored to original state?
* [ ] Have you used non-shell/command modules for Kubernetes tasks?
* [ ] Have you (jinja) templatized custom scripts that is run by the litmusbook, if any? 
* [ ] Have you (jinja) templatized Kubernetes deployment manifests used by the litmusbook, if any?
* [ ] Have you reused/created util functions instead of repeating tasks in the litmusbook?
* [ ] Do the artifacts follow the appropriate directory structure? 
* [ ] Have you isolated storage (eg: OpenEBS) specific implementations, checks? 
* [ ] Have you isolated platform (eg: baremetal kubeadm/openshift/aws/gcloud) specific implementations, checks?  
* [ ] Are the ansible facts well defined? Is the scope explicitly set for playbook & included utils?
* [ ] Have you ensured minimum/careful usage of shell utilities (awk, grep, sed, cut, xargs etc.,)?
* [ ] Can the limtusbook be executed both from within & outside a container (configurable paths, no hardcode)?
* [ ] Can you suggest the minimal resource requirements for the litmusbook execution?
* [ ] Does the litmusbook job artifact carry comments/default options/range for the ENV tunables?
* [ ] Has the litmusbooks been linted? 

**Special notes for your reviewer**:

PR #1779, opened 2020-07-23
chore(litmus-portal): Added Self-Deployer

### LINKED ISSUE #1767: Cluster Self Registration
Automatically deploy all Litmus Experiment related components and register the Red Cluster too in the Litmus Portal.

gdsoumya:

- Adds Self Deployer for Cluster Self-registration
- Sample Self-Deployer manifest for reference

fixes #1767 

**Checklist**


litmus-portal/backend/self-deployer/pkg/k8s/cluster.go
```diff
@@ -0,0 +1,95 @@
...
+	if mapping.Scope.Name() == meta.RESTScopeNameNamespace {
+		// namespaced resources should specify the namespace
+		dr = dyn.Resource(mapping.Resource).Namespace(obj.GetNamespace())
+	} else {
+		// for cluster-wide resources
+		dr = dyn.Resource(mapping.Resource)
+	}
+
+	return applyRequest(obj, requestType, dr)
+}
```
rajdas98, line 95: As these are redundant functions from the agent you can import the package using 
`
go get -v github.com/litmuschaos/litmus/litmus-portal/backend/subscriber/agent/pkg/cluster@litmus-portal-v1
`

Reuse those function.

litmus-portal/backend/self-deployer/pkg/util/uitls.go
```diff
@@ -0,0 +1,101 @@
...
+		if err != nil || res.StatusCode != 200 {
+			time.Sleep(30 * time.Second)
+		} else {
+			break
+		}
+	}
+}
+
+func RegSelfCluster(server, pid string) (string, error) {
+	var jsonStr = []byte(`{"query":"mutation { userClusterReg(clusterInput: { cluster_name: \"Self-Cluster\", description:\"Self-Cluster\", project_id:\"` + pid + `\", platform_name:\"others\", cluster_type:\"INTERNAL\"})}"}`)
```
rajdas98, line 33: Put the graphql query in a string var and remove the escape character to look like a normal string. So that it will be easy to understand the query.

litmus-portal/backend/self-deployer/pkg/util/uitls.go
```diff
@@ -0,0 +1,101 @@
...
+	return nil
+}
+
+func CleanUp(path string) error {
+	body, err := ioutil.ReadFile(path)
+	if err != nil {
+		log.Print(err.Error())
+		return err
+	}
+	response, err := k8s.ClusterOperations(body, "delete")
```
rajdas98, line 90: (Same comment) reuse by importing from agent

litmus-portal/self-deployer-manifest.yml
```diff
@@ -0,0 +1,46 @@
+apiVersion: v1
```
rajdas98, line 1: Why this manifest is redundant?

litmus-portal/backend/self-deployer/build/Dockerfile
```diff
@@ -0,0 +1,17 @@
+FROM golang:1.14 AS builder
+
```
rajdas98, line 2: Use 
`LABEL maintainer="LitmusChaos"`

litmus-portal/backend/self-deployer/build/Dockerfile
```diff
@@ -0,0 +1,17 @@
...
+ADD . /self-deployer
+WORKDIR /self-deployer
+RUN CGO_ENABLED=0 go build -o /output/deployer -v
```
rajdas98, line 5: Do we need CGO_ENABLE=0?

litmus-portal/backend/self-deployer/build/Dockerfile
```diff
@@ -0,0 +1,17 @@
...
+
+
+FROM alpine:latest
+
+RUN apk update && apk add curl bash
+RUN apk add --no-cache libc6-compat
```
rajdas98, line 11: Do we need these commands?

litmus-portal/backend/self-deployer/deployer.go
```diff
@@ -0,0 +1,35 @@
...
+	"log"
+	"os"
+	"strings"
+	"time"
+
+	"github.com/litmuschaos/litmus/litmus-portal/backend/self-deployer/pkg/util"
+)
+
+func main() {
+	active := os.Getenv("SELF_CLUSTER")
```
rajdas98, line 13: +1

litmus-portal/backend/self-deployer/deployer.go
```diff
@@ -0,0 +1,35 @@
...
+	if strings.ToLower(active)=="true" {
```
rajdas98, line 14: +1

rajdas98:
Other than those comments, This PR looks good. Thanks for this PR. 
Maybe it will take some time for the final review. I will try to merge this PR to unblock other things.

litmus-portal/backend/self-deployer/pkg/util/uitls.go
```diff
@@ -0,0 +1,101 @@
...
+	"strings"
+	"time"
+
+	"github.com/litmuschaos/litmus/litmus-portal/backend/self-deployer/pkg/k8s"
+
+)
+
+func WaitForServer(server string) {
+	log.Print("Checking if Server Up ", server)
+	for true {
```
rajdas98, line 19: Will this going to be an infinite loop if the server responds with 200? Maybe you can use a timeout instead of a loop.
Say if the server didn't come up than the self-deployer pod will go `crashbackloop `state, so the user will get to know about the problem. Otherwise, the pod will be in running state.

litmus-portal/backend/self-deployer/pkg/k8s/cluster.go
gdsoumya: Yes I tried, bch is giving warnings but I cannot do that unless it is tagged and published on github first.

litmus-portal/backend/self-deployer/pkg/util/uitls.go
gdsoumya: The escape characters is necessary, it throws errors without those

litmus-portal/backend/self-deployer/pkg/util/uitls.go
gdsoumya: Same as above, cannot reuse unless tagged and published on github

litmus-portal/self-deployer-manifest.yml
gdsoumya: This manifest is not redundant but, I have kept it separate from the k8s-manifest as it will clutter the k8s-manifest.

litmus-portal/backend/self-deployer/build/Dockerfile
gdsoumya: Yes for using a minimal alpine image for the go binary run I have found CGO_ENABLE=0 is a option that is required, or else some libraries might not work on alpine.

litmus-portal/backend/self-deployer/build/Dockerfile
gdsoumya: Yes these are also required for running some go compiled go libraries in alpine image

litmus-portal/backend/self-deployer/pkg/util/uitls.go
gdsoumya: Currently what I am doing is that the self-deployer will loop and check if the GQL server responds 200 or not. If it doesn't I sleep for 30sec and then try again. It is true that it will keep doing this forever if the server doesn't come up, but most probably the server will come up and it will exit the loop, or else the user will notice that there's a error with the portal and debug the deployments. I can try applying some timeout too, but what would be a good timeout period? Like it may happen the server takes some time to come up depending on the resources available.

litmus-portal/backend/self-deployer/pkg/k8s/cluster.go
gdsoumya: Cannot reuse the function as my one is different from this one, I am getting the data as an array of bytes directly, I can maybe reuse the applyRequest but isn't exported.

litmus-portal/backend/self-deployer/pkg/util/uitls.go
gdsoumya: Same reason as above, my function takes different format of data. It might be a little extra work to convert it to your format. Maybe we can keep this as a more generic function as it takes an array of bytes and the request type, we can add a util function to convert from json in go-agent and use these function in there instead?

litmus-portal/backend/self-deployer/pkg/k8s/cluster.go
rajdas98: You have to use ClusterOperation function
Cluster operation function will call applyRequesr

rajdas98: You have to send the data in this format
{
   requesttype: "create",
   manifest: "",
   resourcettype: "",
  namespace: ""
}

gdsoumya: Yes we can maybe reuse this package in go-agent, this function is more generic and it might be easier to use?

litmus-portal/backend/self-deployer/pkg/util/uitls.go
rajdas98: You have to convert string to escape string , before passing to the function

litmus-portal/backend/self-deployer/build/Dockerfile
rajdas98: CGO is used to compile c code in golang. Try to use it without that

litmus-portal/backend/self-deployer/build/Dockerfile
rajdas98: curl and bash is not used in any part in thr program.

rajdas98: Not sure about the lib6-compat

litmus-portal/backend/self-deployer/build/Dockerfile
gdsoumya: I faced issues while doing this and this is one of the common solutions : [StackOverflow](https://stackoverflow.com/questions/36279253/go-compiled-binary-wont-run-in-an-alpine-docker-container-on-ubuntu-host)

litmus-portal/backend/self-deployer/pkg/util/uitls.go
rajdas98: Having a timeout will be good to make the pod crashbackloop to inform  the user about the problem.

rajdas98: Max timeout can be 5 mins

litmus-portal/backend/self-deployer/build/Dockerfile
gdsoumya: Bash is required to run the executable properly. I had faced some problems without bash. Curl can be removed. I'll try without bash and check too.

litmus-portal/backend/self-deployer/template/self-deployer.yml
```diff
@@ -0,0 +1,27 @@
+apiVersion: apps/v1
+kind: Deployment
```
rajdas98, line 2: How this manifest different from self-deployer-manifest.yml?

gdsoumya: This is used by client-go to just delete the self-deployer and not the roles. We can reuse the roles for other purpose that's why it is here.

litmus-portal/backend/self-deployer/pkg/k8s/cluster.go
rajdas98: Yes

litmus-portal/backend/self-deployer/pkg/util/uitls.go
gdsoumya: Okay I'll try adding that, I'll set a panic when the timeout happens. that should give a crashbackloop

litmus-portal/backend/self-deployer/template/self-deployer.yml
rajdas98: We can put the cluster-admin in k8s manifest and  then you can reuse that self-deployer-manifest.yml. 
And where is subscriber manifest(agent + igc)?

gdsoumya: The manifest for the igc+agent is present in the graphql-server/template which is in the other [PR](https://github.com/litmuschaos/litmus/pull/1771). Yes we if we shift the roles we can maybe use it but building the docker image might become a little difficult as we'll have to build the image from the litmus-portal dir as we cannot access files outside the context dir and if we change the context dir, it might increase the context size and slow down the build process.

litmus-portal/k8s-manifest.yml
```diff
@@ -146,3 +146,50 @@ spec:
       targetPort: 27017
   selector:
     component: database
+---
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: deployer-account
```
rajdas98, line 153: Generic service-account

litmus-portal/backend/self-deployer/pkg/k8s/cluster.go
```diff
@@ -0,0 +1,95 @@
...
+func ClusterOperations(manifest []byte, requestType string) (*unstructured.Unstructured, error) {
+	cfg, err := rest.InClusterConfig()
+	if err != nil {
+		log.Printf("err: %v\n", err)
+	}
+
+	dc, err := discovery.NewDiscoveryClientForConfig(cfg)
+	if err != nil {
+		log.Printf("err: %v\n", err)
+	}
```
rajdas98, line 62: Move to a pkg and add out cluster configuration also

PR #1782, opened 2020-07-25
chore(litmus-portal): Refactoring agent to call graphql server directly

rajdas98:

Changes
- Refactoring few agent functionalities to use graphql server directly
- Now the agent is the subscriber (Migrate the agent files to subscriber to dir)
- Adding OutCluster k8s client to run the subscriber directly.
- Removing docker image of agent and igc from the Makefile and adding a new subscriber image to the Makefile
- Refactoring the Dockerfile

Special Note:
- We are deprecating IGC, so removal of igc will go under another PR.
- This PR is dedicated to the agent and its changes.


litmus-portal/backend/subscriber/subscriber.go
```diff
@@ -0,0 +1,99 @@
...
+	req.Header.Set("Content-Type", "application/json")
+	req.Header.Set("Accept", "application/json")
+	req.Header.Set("Connection", "keep-alive")
+	req.Header.Set("DNT", "1")
+	req.Header.Set("Origin", "http://localhost:8080")
+
+	for {
+		resp, err := client.Do(req)
+		if err != nil {
+			log.Fatal(err)
```
rajdas98, line 67: This has to be a HTTP call not a websocket, because URL package is not contain any query parameter. https://godoc.org/net/url#URL

And we don't have to worry about the close connection because its a API call not a database connection.

rajdas98: cc @gdsoumya

rajdas98: And this is not an infinite loop, as it waits for the response. After it gets a response from the graphql server then it goes into a second loop

litmus-portal/backend/subscriber/subscriber.go
```diff
@@ -0,0 +1,99 @@
...
+	"strings"
+)
+
+var GRAPHQL_SERVER_ADDRESS = os.Getenv("SERVER_ADDRESS") // Format: http://IP:PORT/query
+
+var kubeconfig *string
+
+func init() {
+	if GRAPHQL_SERVER_ADDRESS == "" {
+		panic("GRAPHQL SERVER ADDRESS NOT FOUND")
```
rajdas98, line 39: Use of panic is necessary to make the pod into crashbackloop if the address is not present.

rajdas98:
Merging this PR and adding the cluster registration, some refactoring on graphql server will send in a separate PR

rajdas98:
This PR has some frontend side changes because of `npm format`.

litmus-portal/backend/subscriber/subscriber.go
gdsoumya: But it will get some error or timeout as the server won't respond every time?

gdsoumya: Does the request not timeout on the client side? The server will have an open connection but go-client should time out after some time right?

rajdas98: If the server goes down, it will automatically timeout

PR #1815, opened 2020-08-03
litmus-portal(): Refactoring in GQL Server

rajdas98:
Changes-
1. Changing GetCluster function to use findOne operation instead of find
2. Generalize some function
3. Refactor the directory structure.
4. Changing the name of some env variable to some meaningful names
5. Generating resolvers for the Workflow schema
6. Adding a function to instantiate stateData struct 
    ref- https://github.com/litmuschaos/litmus/blob/litmus-portal/litmus-portal/backend/graphql-server/pkg/data-store/store.go#L10
7. Separating mongodb struct from gql struct.
8. Making UpdateCluster and UpdateClusterData more generic.


litmus-portal/backend/graphql-server/graph/model/models_gen.go
```diff
@@ -3,17 +3,17 @@
...
+	ProjectID    string  `bson:"project_id" bson:"project_id"`
+	ClusterName  string  `bson:"cluster_name" bson:"cluster_name"`
+	Description  *string `bson:"description" bson:"description"`
+	PlatformName string  `bson:"platform_name" bson:"platform_name"`
+	AccessKey    string  `bson:"access_key" bson:"access_key"`
+	IsRegistered bool    `bson:"is_registered" bson:"is_registered"`
+	IsActive     bool    `bson:"is_active" bson:"is_active"`
+	UpdatedAt    string  `bson:"updated_at" bson:"updated_at"`
+	CreatedAt    string  `bson:"created_at" bson:"created_at"`
+	ClusterType  string  `bson:"cluster_type" bson:"cluster_type"`
```
rajdas98, line 16: Need to change json to bson for mongodb

litmus-portal/backend/graphql-server/pkg/data-store/store.go
```diff
@@ -13,11 +13,10 @@ type StateData struct {
...
-	State.ConnectedCluster = make(map[string]chan *model.ClusterAction)
-	State.Mutex = &sync.Mutex{}
-}
+func New() *StateData {
+	return &StateData{
+		ClusterEventPublish: make(map[string][]chan *model.ClusterEvent),
+		ConnectedCluster: make(map[string]chan *model.ClusterAction),
+		Mutex: &sync.Mutex{},
+	}
+}
```
rajdas98, line 22: We can call the New() to instantiate the structure

litmus-portal/backend/graphql-server/pkg/database/mongodb/mongodb.go
```diff
@@ -13,95 +13,82 @@ import (
 )
 
 var clusterCollection *mongo.Collection
-var backgroundContext context.Context
+var backgroundContext = context.Background()
```
rajdas98, line 16: Global context should be reuse by every function

litmus-portal/backend/graphql-server/pkg/database/mongodb/mongodb.go
```diff
@@ -13,95 +13,82 @@ import (
...
+	dbServer := os.Getenv("MONGODB_SERVER")
 	dbName := os.Getenv("DB_NAME")
 	collectionName := os.Getenv("COLLECTION_NAME")
-	clientOptions := options.Client().ApplyURI("mongodb://" + dbServer)
-	client, err := mongo.Connect(context.TODO(), clientOptions)
 
+	clientOptions := options.Client().ApplyURI("mongodb://" + dbServer)
+	client, err := mongo.Connect(backgroundContext, clientOptions)
 	if err != nil {
-		log.Fatal(err)
```
rajdas98, line 27: Need to print the error in some specific area. Otherwise, return the error

litmus-portal/backend/graphql-server/pkg/file_handlers/file_handlers.go
```diff
@@ -0,0 +1,84 @@
+package file_handlers
```
rajdas98, line 1: I move this file from handler/handler.go to file_handlers/file_handlers.go because it creates some confusion between HTTP pkg which is also using the same dir structure.

litmus-portal/backend/graphql-server/pkg/file_handlers/file_handlers.go
```diff
@@ -0,0 +1,84 @@
...
+	"os"
+	"strings"
+
+	"github.com/gorilla/mux"
+	"github.com/litmuschaos/litmus/litmus-portal/backend/graphql-server/pkg/cluster"
+	"github.com/litmuschaos/litmus/litmus-portal/backend/graphql-server/pkg/database/mongodb"
+)
+
+//ManifestParser parses manifests yaml and generates dynamic manifest with specified keys
+func manifestParser(id, key, server, template string) ([]string, error) {
```
rajdas98, line 17: I move the manifestParse() from pkg/cluster because having one function in a file doesn't help.
This is the right place where we can use the manifestParser()

litmus-portal/backend/graphql-server/pkg/graphql/mutations/mutation.go
```diff
@@ -0,0 +1,85 @@
+package mutations
```
rajdas98, line 1: I moved all the mutation related functions from pkg/cluster to pkg/graphql/mutations, because every file contains only one function and again single function in a file doesn't make sense.
If this file grows, we need to think for the modularity of mutation.go

litmus-portal/backend/graphql-server/pkg/graphql/subscriptions/subscription.go
```diff
@@ -1,4 +1,4 @@
-package cluster
+package subscriptions
```
rajdas98, line 1: Same reason-  https://github.com/litmuschaos/litmus/pull/1815/files#r464629072

litmus-portal/backend/graphql-server/server.go
```diff
@@ -1,30 +1,44 @@
...
+	}
+
+	if err = database.DBInit(); err != nil {
+		log.Fatal(err)
+	}
+}
+
 func main() {
-	port := defaultPort
-	database.DBInit()
```
rajdas98, line 21: Earlier it is not checking the err return from the DBInit Function

litmus-portal/backend/graphql-server/graph/model/models_gen.go
gdsoumya: It is already there, the json is for gql and bson is for mongodb. There are 2 tags for the Cluster type.

litmus-portal/backend/graphql-server/pkg/database/mongodb/mongodb.go
gdsoumya: It is being reused, the context is being set to context.Background() during the DBInit() call. It is responsible to initialize all DB related data.

litmus-portal/backend/graphql-server/pkg/database/mongodb/mongodb.go
gdsoumya: This is a fatal error, so we need to panic in this case, we can do this right here or handle it in the main func.

litmus-portal/backend/graphql-server/pkg/graphql/mutations/mutation.go
gdsoumya: Yes I made it in separate files for the ease of locating the particular function, but having it in one file makes sense.

litmus-portal/backend/graphql-server/server.go
gdsoumya: Yes, I was throwing an error inside that function itself.

rajdas98: We have to throw an error at a single place, as the other functions are returning the error.

litmus-portal/backend/graphql-server/pkg/database/mongodb/mongodb.go
rajdas98: https://github.com/litmuschaos/litmus/pull/1815#discussion_r465891708

litmus-portal/backend/graphql-server/pkg/database/mongodb/mongodb.go
rajdas98: It is not being reused in this place https://github.com/litmuschaos/litmus/blob/litmus-portal/litmus-portal/backend/graphql-server/pkg/database/db.go#L24

litmus-portal/backend/graphql-server/graph/model/models_gen.go
rajdas98: If you don't convert it to BSON, it will remove underscore while inserting data into the database
ex- updated_at will `updatedat` in db.

Check the mongodb after inserting the data

litmus-portal/backend/graphql-server/pkg/database/mongodb/mongodb.go
```diff
@@ -0,0 +1,93 @@
...
+func InsertCluster(cluster model.Cluster) error {
+	ctx, _ := context.WithTimeout(backgroundContext, 10*time.Second)
+	_, err := clusterCollection.InsertOne(ctx, cluster)
+	if err != nil {
+		return err
+	}
+	return nil
+}
+
+func GetCluster(cluster_id string) (model.Cluster, error) {
```
rajdas98, line 56: Changing Find to FindOne operation

litmus-portal/backend/graphql-server/pkg/graphql/mutations/mutation.go
rajdas98: If we put one function in a single file, the files inside a directory will grow. which is not a good approach, we need to generalize the functions as much as possible.

litmus-portal/backend/graphql-server/graph/model/models_gen.go
rajdas98: this structure is also used by mongodb while inserting and retrieving the data

rajdas98: We need to separate MongoDB struct and gql struct because if someones generate the code again bson will be removed automatically.

gdsoumya: You have changed the tag to just bson, in the previous version I kept both json and bson tags for Cluster type because it is required by both GQL and MongoDB. I believe GQL requires json tags, that's why I had kept both the tags. You can take a look at the older version of this file.

litmus-portal/backend/graphql-server/pkg/database/mongodb/mongodb.go
gdsoumya: Yes that one I had kept as context.TODO, we can change that to use this one, we have to set the variable before that, currently we are initializing it after line 24.

litmus-portal/backend/graphql-server/pkg/database/mongodb/mongodb.go
gdsoumya: You need to change the util functions too in the cluster package also.

litmus-portal/backend/graphql-server/graph/model/models_gen.go
gdsoumya: I was thinking of separating the two models but the structures are same and as the collections increase a lot of code will be repeated,

litmus-portal/backend/graphql-server/pkg/database/mongodb/mongodb.go
```diff
@@ -0,0 +1,93 @@
...
+var backgroundContext = context.Background()
+var err error
+
+var collection = map[string]string{
+	"Cluster": "cluster-collection",
+}
+
+var dbName = "litmus"
+
+type Cluster struct {
```
rajdas98, line 24: Seperating the mongodb struct from gql struct

litmus-portal/backend/graphql-server/pkg/database/mongodb/mongodb.go
```diff
@@ -0,0 +1,93 @@
...
+	var cluster Cluster
+	err = clusterCollection.FindOne(ctx, query).Decode(&cluster)
+	if err != nil {
+		return Cluster{}, err
+	}
+
+	return cluster, nil
+}
+
+func UpdateCluster(query bson.D, update bson.D) error {
```
rajdas98, line 84: Making UpdateClusterData and UpdateCluster function generic.

litmus-portal/backend/graphql-server/graph/model/models_gen.go
rajdas98: I am separating mongodb struct from gql struct in this PR

litmus-portal/backend/graphql-server/pkg/database/mongodb/mongodb.go
rajdas98: Already changed in this PR

litmus-portal/backend/graphql-server/pkg/database/mongodb/mongodb.go
gdsoumya: Not sure if we should generalize all functions, surely it will reduce the amount of code. But according to me specialized functions are much easier to debug and also help restrict unwanted changes.

rajdas98: Yes, sometimes it is hard to debug. But we have to reduce the reductant code as much as possible.

litmus-portal/backend/graphql-server/server.go
```diff
@@ -1,30 +1,44 @@
...
+		port = defaultPort
+	}
+
 	srv := handler.NewDefaultServer(generated.NewExecutableSchema(generated.Config{Resolvers: &graph.Resolver{}}))
 	router := mux.NewRouter()
 	router.Handle("/", playground.Handler("GraphQL playground", "/query"))
 	router.Handle("/query", srv)
-	router.HandleFunc("/file/{key}", handlers.FileHandler)
-	log.Println("Server is running")
+	router.HandleFunc("/file/{key}{path:.yaml}", file_handlers.FileHandler)
```
rajdas98, line 40: Adding .yaml path

rajdas98: Now, subscriber manifest will be accessible though, http://IP:PORT/file/RandomString.yaml

rajdas98:
I am going to merge this PR after the circleci build passed. cc @gdsoumya

PR #1841, opened 2020-08-07
chore(litmus-portal): Added Workflow Agent Operations to GQL Server

gdsoumya:

## Proposed changes

This PR adds the required mutations, queries, subscriptions and db operations for workflow-agent

## Types of changes

## Checklist

### DIFF
--- a/litmus-portal/backend/graphql-server/graph/model/models_gen.go
+++ b/litmus-portal/backend/graphql-server/graph/model/models_gen.go
@@ -17,6 +17,17 @@ type Cluster struct {
 	ClusterType        string  `json:"cluster_type"`
 }
 
+type WorkflowRun struct {
+	WorkflowRunID string `json:"workflow_run_id" bson:"workflow_run_id"`
+	WorkflowID    string `json:"workflow_id" bson:"workflow_id"`
+	ClusterName   string `json:"cluster_name" bson:"cluster_name"`
+	LastUpdated   string `json:"last_updated" bson:"last_updated"`
+	ProjectID     string `json:"project_id" bson:"project_id"`
+	ClusterID     string `json:"cluster_id" bson:"cluster_id"`
+	WorkflowName  string `json:"workflow_name" bson:"workflow_name"`
+	ExecutionData string `json:"execution_data" bson:"execution_data"`
+}
+
 type ClusterAction struct {
 	ProjectID string `json:"project_id"`
 	Action    string `json:"action"`
@@ -59,6 +70,40 @@ type ClusterInput struct {
 	PlatformName string  `json:"platform_name"`
 	ProjectID    string  `json:"project_id"`
 	ClusterType  string  `json:"cluster_type"`
+}
+
+type PodLog struct {
+	ClusterID     *ClusterIdentity `json:"cluster_id"`
+	RequestID     string           `json:"request_id"`
+	WorkflowRunID string           `json:"workflow_run_id"`
+	PodName       string           `json:"pod_name"`
+	PodType       string           `json:"pod_type"`
+	Log           string           `json:"log"`
+}
+
+type PodLogRequest struct {
+	ClusterID      string  `json:"cluster_id"`
+	WorkflowRunID  string  `json:"workflow_run_id"`
+	PodName        string  `json:"pod_name"`
+	PodNamespace   string  `json:"pod_namespace"`
+	PodType        string  `json:"pod_type"`
+	ExpPod         *string `json:"exp_pod"`
+	RunnerPod      *string `json:"runner_pod"`
+	ChaosNamespace *string `json:"chaos_namespace"`
+}
+
+type PodLogResponse struct {
+	WorkflowRunID string `json:"workflow_run_id"`
+	PodName       string `json:"pod_name"`
+	PodType       string `json:"pod_type"`
+	Log           string `json:"log"`
+}
+
+type WorkflowRunInput struct {
+	WorkflowRunID string           `json:"workflow_run_id"`
+	WorkflowName  string           `json:"workflow_name"`
+	ExecutionData string           `json:"execution_data"`
+	ClusterID     *ClusterIdentity `json:"cluster_id"`
 }
 
 type ChaosWorkFlowInput struct {
--- a/litmus-portal/backend/graphql-server/graph/model/models_gen.go
+++ b/litmus-portal/backend/graphql-server/graph/model/models_gen.go
@@ -18,14 +18,14 @@ type Cluster struct {
 }
 
 type WorkflowRun struct {
-	WorkflowRunID string `json:"workflow_run_id" bson:"workflow_run_id"`
-	WorkflowID    string `json:"workflow_id" bson:"workflow_id"`
-	ClusterName   string `json:"cluster_name" bson:"cluster_name"`
-	LastUpdated   string `json:"last_updated" bson:"last_updated"`
-	ProjectID     string `json:"project_id" bson:"project_id"`
-	ClusterID     string `json:"cluster_id" bson:"cluster_id"`
-	WorkflowName  string `json:"workflow_name" bson:"workflow_name"`
-	ExecutionData string `json:"execution_data" bson:"execution_data"`
+	WorkflowRunID string `json:"workflow_run_id"`
+	WorkflowID    string `json:"workflow_id"`
+	ClusterName   string `json:"cluster_name"`
+	LastUpdated   string `json:"last_updated"`
+	ProjectID     string `json:"project_id"`
+	ClusterID     string `json:"cluster_id"`
+	WorkflowName  string `json:"workflow_name"`
+	ExecutionData string `json:"execution_data"`
 }
 
 type ClusterAction struct {
--- a/litmus-portal/backend/graphql-server/pkg/cluster/cluster_jwt.go
+++ b/litmus-portal/backend/graphql-server/pkg/cluster/cluster_jwt.go
@@ -9,7 +9,7 @@ import (
 	"github.com/dgrijalva/jwt-go"
 )
 
-var secret = os.Getenv("CLUSTER_JWT_SECRET")
+var secret = os.Getenv("JWT_SECRET")
 
 //ClusterCreateJWT generates jwt used in cluster registration
 func ClusterCreateJWT(id string) (string, error) {
--- a/litmus-portal/backend/graphql-server/pkg/database/mongodb/init.go
+++ b/litmus-portal/backend/graphql-server/pkg/database/mongodb/init.go
@@ -15,7 +15,7 @@ var workflowRunCollection *mongo.Collection
 var backgroundContext = context.Background()
 var err error
 
-var collection = map[string]string{
+var collections = map[string]string{
 	"Cluster":     "cluster-collection",
 	"WorkflowRun": "workflow-run",
 }
@@ -37,6 +37,17 @@ type Cluster struct {
 	ClusterType        string  `bson:"cluster_type"`
 }
 
+type WorkflowRun struct {
+	WorkflowRunID string `bson:"workflow_run_id"`
+	WorkflowID    string `bson:"workflow_id"`
+	ClusterName   string `bson:"cluster_name"`
+	LastUpdated   string `bson:"last_updated"`
+	ProjectID     string `bson:"project_id"`
+	ClusterID     string `bson:"cluster_id"`
+	WorkflowName  string `bson:"workflow_name"`
+	ExecutionData string `bson:"execution_data"`
+}
+
 //DBInit initializes database connection
 func DBInit() error {
 	dbServer := os.Getenv("DB_SERVER")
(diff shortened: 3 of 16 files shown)


litmus-portal/backend/graphql-server/pkg/cluster/cluster_jwt.go
```diff
@@ -9,7 +9,7 @@ import (
 	"github.com/dgrijalva/jwt-go"
 )
 
-var secret = os.Getenv("CLUSTER_JWT_SECRET")
+var secret = os.Getenv("JWT_SECRET")
```
rajdas98, line 12: It will create confusion with auth JWT

litmus-portal/backend/graphql-server/graph/model/models_gen.go
```diff
@@ -17,6 +17,17 @@ type Cluster struct {
...
+	WorkflowRunID string `json:"workflow_run_id" bson:"workflow_run_id"`
+	WorkflowID    string `json:"workflow_id" bson:"workflow_id"`
+	ClusterName   string `json:"cluster_name" bson:"cluster_name"`
+	LastUpdated   string `json:"last_updated" bson:"last_updated"`
+	ProjectID     string `json:"project_id" bson:"project_id"`
+	ClusterID     string `json:"cluster_id" bson:"cluster_id"`
+	WorkflowName  string `json:"workflow_name" bson:"workflow_name"`
+	ExecutionData string `json:"execution_data" bson:"execution_data"`
+}
+
```
rajdas98, line 30: You can seperate the struct for bson to mongo.go and convert mongo struct to gql struct whenever required.
Ref- clustercreate struct

litmus-portal/backend/graphql-server/pkg/graphql/mutations/mutation.go
```diff
@@ -88,3 +88,50 @@ func NewEvent(clusterEvent model.ClusterEventInput, r store.StateData) (string,
...
+			PodName:       podLog.PodName,
+			WorkflowRunID: podLog.WorkflowRunID,
+			PodType:       podLog.PodType,
+			Log:           podLog.Log,
+		}
+		reqChan <- &resp
+		close(reqChan)
+		return "LOGS SENT SUCCESSFULLY", nil
+	}
+	return "LOG REQUEST CANCELLED", nil
```
rajdas98, line 136: This return will throw an error, but it is nil. Make to one parameter response

litmus-portal/backend/graphql-server/pkg/graphql/mutations/mutation.go
```diff
@@ -88,3 +88,50 @@ func NewEvent(clusterEvent model.ClusterEventInput, r store.StateData) (string,
...
+		ExecutionData: input.ExecutionData,
+		WorkflowID:    "000000000000",
+	}
+	subscriptions.SendWorkflowEvent(newWorkflowRun, r)
+	err = database.UpsertWorkflowRun(newWorkflowRun)
+	if err != nil {
+		log.Print("ERROR", err)
+		return "", err
+	}
+	return "Workflow Run Accepted", nil
```
rajdas98, line 115: https://github.com/litmuschaos/litmus/pull/1841/files#r467488940

litmus-portal/backend/graphql-server/pkg/cluster/cluster_jwt.go
gdsoumya: It has to be the same as the auth server? else we won't be able to verify the user sessions in gql too.

litmus-portal/backend/graphql-server/pkg/graphql/mutations/mutation.go
```diff
@@ -88,3 +88,50 @@ func NewEvent(clusterEvent model.ClusterEventInput, r store.StateData) (string,
...
+		ClusterName:   cluster.ClusterName,
+		ProjectID:     cluster.ProjectID,
+		LastUpdated:   strconv.FormatInt(time.Now().Unix(), 10),
+		WorkflowRunID: input.WorkflowRunID,
+		WorkflowName:  input.WorkflowName,
+		ExecutionData: input.ExecutionData,
+		WorkflowID:    "000000000000",
+	}
+	subscriptions.SendWorkflowEvent(newWorkflowRun, r)
+	err = database.UpsertWorkflowRun(newWorkflowRun)
```
rajdas98, line 110: This function UpsertWorkflowRun will create confusion because of the same naming convention

litmus-portal/backend/graphql-server/pkg/graphql/mutations/mutation.go
gdsoumya: We need both a string return and a error return because the output is used by the GQL resolver which requires a string(response) and error

litmus-portal/backend/graphql-server/pkg/graphql/mutations/mutation.go
gdsoumya: I'll rename it to WorkflowRunHandler and the database func can remain the same

litmus-portal/backend/graphql-server/pkg/graphql/mutations/mutation.go
rajdas98: You have to throw an error here

litmus-portal/backend/graphql-server/pkg/cluster/cluster_jwt.go
rajdas98: There will be two JWT
1. CLUSTER_JWT_SECET
2. AUTH_JWT_SECRET.

litmus-portal/backend/graphql-server/pkg/graphql/mutations/mutation.go
gdsoumya: No it is not an error, so basically the frontend can cancel the log request anytime they want as it is on-demand and throwing an error here isn't required as the agent doesn't need to know about this.

litmus-portal/backend/graphql-server/pkg/graphql/queries/query.go
```diff
@@ -0,0 +1,46 @@
...
+//GetWorkflowRuns sends all the workflow runs for a project from the DB
+func QueryWorkflowRuns(pid string) ([]*model.WorkflowRun, error) {
+	wfRuns, err := database.GetWorkflowRuns(pid)
+	if err != nil {
+		return nil, err
+	}
+	result := []*model.WorkflowRun{}
+	for i := 0; i < len(wfRuns); i++ {
+		result = append(result, &wfRuns[i])
+	}
```
rajdas98, line 20:
```suggestion
	for wfrun := range wfRuns {
	  		result = append(result, &wfRuns[wfrun])
	}
```

litmus-portal/backend/graphql-server/pkg/cluster/cluster_jwt.go
gdsoumya: The value of the 2 will have to be same, that's why I thought if keeping it same in both servers

litmus-portal/backend/graphql-server/pkg/graphql/queries/query.go
```diff
@@ -0,0 +1,46 @@
...
+	return result, nil
+}
+
+//GetLogs query is used to fetch the logs from the cluster
+func GetLogs(reqID string, pod model.PodLogRequest, r store.StateData) {
+	data, err := json.Marshal(pod)
+	if err != nil {
+		log.Print("ERROR WHILE MARSHALLING POD DETAILS")
+	}
+	payload := model.ClusterAction{
```
rajdas98, line 30: https://github.com/litmuschaos/litmus/pull/1841/files#r467488791

litmus-portal/backend/graphql-server/pkg/graphql/mutations/mutation.go
```diff
@@ -88,3 +88,50 @@ func NewEvent(clusterEvent model.ClusterEventInput, r store.StateData) (string,
...
 }
+
+// UpsterWorkFlowRun Updates or Inserts a new Workflow Run into the DB
+func UpsterWorkFlowRun(input model.WorkflowRunInput, r store.StateData) (string, error) {
+	cluster, err := cluster.VerifyCluster(*input.ClusterID)
+	if err != nil {
+		log.Print("ERROR", err)
+		return "", err
+	}
+	newWorkflowRun := model.WorkflowRun{
```
rajdas98, line 99: https://github.com/litmuschaos/litmus/pull/1841/files#r467488791

litmus-portal/backend/graphql-server/pkg/cluster/cluster_jwt.go
rajdas98: It's better to seperate these two JWT, as they both have different usecase

litmus-portal/backend/graphql-server/graph/model/models_gen.go
gdsoumya: I don't think we are having a similar usecase for the Cluster struct where both the mongo and gql structs are required at the same time. In this case I need both at the same time, hence I have to initalize the same thing twice.

litmus-portal/backend/graphql-server/pkg/cluster/cluster_jwt.go
rajdas98: and value will be also different

litmus-portal/backend/graphql-server/pkg/database/mongodb/init.go
```diff
@@ -6,17 +6,18 @@ import (
...
 	"go.mongodb.org/mongo-driver/mongo"
 	"go.mongodb.org/mongo-driver/mongo/options"
 )
 
 var clusterCollection *mongo.Collection
+var workflowRunCollection *mongo.Collection
 var backgroundContext = context.Background()
 var err error
 
 var collection = map[string]string{
```
rajdas98, line 18: > 
```suggestion
var collections = map[string]string{
```
I forgot to make it plural

litmus-portal/backend/graphql-server/pkg/cluster/cluster_jwt.go
gdsoumya: The value cannot be different, because we will need to validate the tokens from the auth server in gql server too. To do that either we have to mention another jwt_secret env var or use the same one for this usecase too.

litmus-portal/backend/graphql-server/graph/model/models_gen.go
rajdas98: Can you point out the places?

litmus-portal/backend/graphql-server/pkg/graphql/mutations/mutation.go
gdsoumya: Yes here I am using the WorkflowRun struct and using the same instance for bth gql and mongo operations at line 109 and 110. If we separate the 2 structs I will have to repeat this block of code again

litmus-portal/backend/graphql-server/graph/model/models_gen.go
gdsoumya: https://github.com/litmuschaos/litmus/pull/1841#discussion_r467490919

rajdas98: We don't have to modify this auto-gen struct. It will remove if we regenerate again.

gdsoumya: I agree that it will be removed once regenerated but as general coding practice we should not be repeating the same code if we can reuse it, the models dir can be reused for mongo too without any changes. The regerate problem is only during dev but once in prod we won't have to regenerate the schema that frequently

litmus-portal/backend/graphql-server/pkg/graphql/mutations/mutation.go
rajdas98: You don't have to repeat the struct. You can simply convert one struct to another.
Ref- https://github.com/litmuschaos/litmus/blob/litmus-portal/litmus-portal/backend/graphql-server/graph/schema.resolvers.go#L95
here I converted verifiedCluster struct to gql struct

litmus-portal/backend/graphql-server/pkg/graphql/mutations/mutation.go
```diff
@@ -88,3 +88,50 @@ func NewEvent(clusterEvent model.ClusterEventInput, r store.StateData) (string,
...
+		ClusterID:     cluster.ClusterID,
+		ClusterName:   cluster.ClusterName,
+		ProjectID:     cluster.ProjectID,
+		LastUpdated:   strconv.FormatInt(time.Now().Unix(), 10),
+		WorkflowRunID: input.WorkflowRunID,
+		WorkflowName:  input.WorkflowName,
+		ExecutionData: input.ExecutionData,
+		WorkflowID:    "000000000000",
+	}
+	subscriptions.SendWorkflowEvent(newWorkflowRun, r)
```
rajdas98, line 109:
```suggestion
	subscriptions.SendWorkflowEvent(model.newWorkflowRun(newWorkflowRun), r)
```
This can be converted like this after you move mongo struct from the gql struct

litmus-portal/backend/graphql-server/pkg/graphql/mutations/mutation.go
gdsoumya: Okay, I'll try to convert it then. But having the same struct twice might not be an efficient way of doing this.

rajdas98: Yes, we can do this if the gql struct is not autogenerated. We don't have to change autogen struct.

PR #1843, opened 2020-08-09
chore(litmus-portal): Adding cluster confirm check with config map and connection of gql with subscriber

rajdas98:
## Proposed changes
- Adding default config map to subscriber's manifest.
- Adding GQL mutation request to confirm cluster and generates a new cluster key
- Adding connection of GQL server with subscriber using a subscription.
- deleting deprecated agent code
- Few more reactors to subscriber.go
- Coverting Dockerfile to multi-stage docker build

## Types of changes

## Checklist

## Dependency

fixes- https://github.com/litmuschaos/litmus/issues/1889


litmus-portal/backend/graphql-server/manifests/subscriber.yml
```diff
@@ -24,3 +24,11 @@ spec:
               value: #{CID}
             - name: KEY
               value: #{KEY}
+---
+apiVersion: v1
```
gdsoumya, line 28: If the manifest is reapplied the configmap data will also be re-configured?

rajdas98: Yes, it will re-configured with this config

gdsoumya: So we cannot have the user reapply this, we need to create this config map only after the subscriber comes up? Because if they unknowingly reapply the whole thing is reconfigured and identity of the cluster changes? We should deploy the subscriber and do the pre-check inside it and if it's a new install we create the configmap through the subscriber directly.

rajdas98: What if the user reapply subscriber, it will also reconfigured the confimap config?

rajdas98: Future work- if user reapplies this subscriber manifest, the subscriber will call the cluster confirmed mutation and if cluster_confimed is true in MongoDB, it will send the new key, which then it will update it in config map.

gdsoumya: Yes it would reconfigure the subscriber but the configmap will remain untouched, so when the subscriber restarts it will get the old details and will be able to verify that the cluster is already registered so it won't re-register the cluster, it'll skip that step and use the old identity.

rajdas98: For now, it's okay to have it here. After that, we need to modify cluster confirm mutation

gdsoumya: The behavior of cluster-confirm is the same as you have stated, if we send a cluster-confirm request again it will update the key. That is not a problem, but as the subscriber will be responsible for installing all the other components if it is restarted and has no context of whether there is already an installation for the same, it will reinstall everything from scratch that might cause some downtime considering a workflow was scheduled right at that moment.

gdsoumya: That would also mean any previous workflow/experiment schedules will be re-configured or removed too.

litmus-portal/backend/subscriber/subscriber.go
```diff
@@ -1,65 +1,78 @@
-/*
```
rajdas98, line 1: We need to write a separate README.md for the subscriber. Removing this for now

litmus-portal/backend/subscriber/pkg/cluster/cluster.go
```diff
@@ -92,25 +123,62 @@ func ClusterOperations(clientData map[string]interface{}, kubeconfig *string) (*
...
+			return nil, err
+		}
+
+		log.Println("Resource successfully applied")
+		return response, nil
+	} else {
+		return nil, fmt.Errorf("err: %v\n", "Invalid Request")
+	}
+
+	return nil, fmt.Errorf("Invalid request")
```
rajdas98, line 183: These lines are moved from applyRequest() function because this function is used only by the clusteroperation() function. also, it will be difficult to understand if we separate these if/else condition

litmus-portal/backend/subscriber/pkg/cluster/k8s.go
```diff
@@ -1,52 +1,64 @@
-package k8s
+package cluster
```
rajdas98, line 1: k8.go is moved from k8s pkg to cluster pkg because to reuse type.go structs which are inside cluster pkg.

rajdas98:
Subscribe logs:
```W0820 15:14:39.634646       1 client_config.go:552] Neither --kubeconfig nor --master was specified.  Using the inClusterConfig.  This might not work.
2020/08/20 15:14:39 cluster confirmed
W0820 15:14:39.651142       1 client_config.go:552] Neither --kubeconfig nor --master was specified.  Using the inClusterConfig.  This might not work.
2020/08/20 15:14:39 Configmap Updated
2020/08/20 15:14:39 Waiting for the subscription...
W0820 15:14:54.931299       1 client_config.go:552] Neither --kubeconfig nor --master was specified.  Using the inClusterConfig.  This might not work.
2020/08/20 15:14:54 Resource successfully applied
2020/08/20 15:14:54 Waiting for the subscription...
```

rajdas98:
Server Logs:
```
2020/08/20 15:10:53 Connected To MONGODB
2020/08/20 15:10:53 connect to http://localhost:8080/ for GraphQL playground
2020/08/20 15:11:19 NEW CLUSTER REGISTERED : ID-3862fbff-2831-482b-8e65-6ebc4ce835bc PID-123
2020/08/20 15:14:39 CLUSTER Confirmed : ID-3862fbff-2831-482b-8e65-6ebc4ce835bc PID-123

```

litmus-portal/backend/subscriber/pkg/cluster/cluster.go
```diff
@@ -2,88 +2,173 @@ package cluster
...
+			Kind:       "ConfigMap",
+			APIVersion: "v1",
+		},
+		ObjectMeta: metav1.ObjectMeta{
+			Name: PortalConfigName,
+		},
+		Data: configMapData,
+	}
+
+	_, err = clientset.CoreV1().ConfigMaps(DefaultNamespace).Update(Ctx, &newConfigMap, metav1.UpdateOptions{})
```
gdsoumya, line 94: Wouldn't it be better to create the configmap in this function rather than creating a dummy one first in the `isClusterConfirmed` function?

gdsoumya (changes requested):

litmus-portal/backend/subscriber/pkg/cluster/cluster.go
rajdas98: Yes, it is possible. Thanks for finding this

rajdas98: Moved to clusterRegister()

litmus-portal/backend/subscriber/build/Dockerfile
```diff
@@ -1,22 +1,19 @@
...
-COPY . .
+LABEL maintainer="LitmusChaos"
+
+COPY --from=builder /output/subscriber /
+CMD ["./subscriber"]
 
-# Go build
-RUN go build -o subscriber .
-ENTRYPOINT ["./subscriber"]
+EXPOSE 8080
```
gdsoumya, line 19: Is this required? Exposed port 8080?

gdsoumya (changes requested):

litmus-portal/backend/subscriber/build/Dockerfile
rajdas98: Not required

gdsoumya (dismissed):
LGTM

litmus-portal/backend/subscriber/subscriber.go
```diff
@@ -1,117 +1,75 @@
...
+			log.Println("cluster confirmed")
+			clusterData["KEY"] = strings.TrimSpace(responseInterface["data"]["clusterConfirm"]["newClusterKey"].(string))
+			cluster.ClusterRegister(clusterData)
+		} else {
+			log.Fatal("Cluster not confirmed")
+		}
+	}
+}
 
+func homeDir() string {
```
gdsoumya, line 64: This one is for out-cluster config right? Not very sure if this will work if someone sets the kubeconfig path to a different dir. What I did was just took the whole kubeconfig path as an env variable.

gdsoumya (changes requested):

litmus-portal/backend/subscriber/subscriber.go
rajdas98: Not we don't have to pass kubeconfig,

rajdas98: Generally, In outcluster, the path /home/user/.kube/config

rajdas98: Outcluster is just for local development

rajdas98: Actually, both is present, if you don't pass kubeconfig. It will take it from default config path. else it will take it from the flag

gdsoumya (approved):
LGTM

PR #1863, opened 2020-08-14
chore(litmus-portal): Adding get cluster details GQL API

rajdas98:

## Proposed changes

- Adding get cluster details GQL API
- Schema addition of getCluster query

## Types of changes

## Checklist

## Dependency

### DIFF
--- a/litmus-portal/backend/graphql-server/graph/schema.graphqls
+++ b/litmus-portal/backend/graphql-server/graph/schema.graphqls
@@ -34,13 +34,13 @@ type ClusterEvent {
 }
 
 type ClusterAction {
-    project_id: ID!
-    action: String!
+  project_id: ID!
+  action: String!
 }
 
 input ClusterActionInput {
-    cluster_id: ID!
-    action:String!
+  cluster_id: ID!
+  action:String!
 }
 
 input ClusterEventInput{
@@ -56,9 +56,9 @@ input ClusterIdentity{
 }
 
 type ClusterConfirmResponse {
-    isClusterConfirmed: Boolean!
-    newClusterKey: String
-    cluster_id: String
+  isClusterConfirmed: Boolean!
+  newClusterKey: String
+  cluster_id: String
 }
 
 input chaosWorkFlowInput {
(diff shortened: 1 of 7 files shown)


litmus-portal/backend/graphql-server/graph/schema.graphqls
```diff
@@ -144,6 +144,8 @@ input PodLogRequest{
 
 type Query{
   getWorkFlowRuns(project_id: String!): [WorkflowRun!]!
+
+  getCluster(project_id: String!, cluster_type: String!): [Cluster!]!
```
gdsoumya, line 148: Is cluster_type needed? the frontend when querying for all the cluster will not need to make 2 queries for this, we can just fetch all the clusters and show them according to the requirements in the frontend.

rajdas98: There is a button called ignite-cluster (internal cluster) to select the target cluster. By clicking on that button an API will be called.

gdsoumya: To get a list of all the cluster we need to get all the clusters internal as well as external, if we have this query we will have to make 2 separate queries. Only after we list the clusters we can select a cluster for the workflow.

gdsoumya: This might work for now but later we will have to change this to accommodate external clusters, the `Cluster` response does have a field called cluster_type which can be used in the frontend to determine whether internal or external.

rajdas98: In the frontend screen, either he  can select an internal cluster or external cluster

rajdas98: Selection_003

rajdas98: internal can be one but there will be many external clusters. If we call all cluster, frontend have to filter it out.

gdsoumya: I got that part, but when we are listing these clusters we have to query the db, so if we specify the cluster_type we have to make 2 queries, 1 for internal and 1 for external. Which doesn't make much sense because there will always be 1 internal cluster. So we can remove that parameter and only query for all the clusters. Then in the frontend we can detect which one is the internal cluster.

rajdas98: we can make the cluster_type not mandatory if cluster_type is not present. it will list all clusters else it will filter based on cluster_type.

gdsoumya: We will have to show all the clusters here right? Because this page is used by the user when they want to create a workflow for a cluster, that cluster can be any cluster either internal or external.

gdsoumya: Yes having it not mandatory would be good, that should solve it.

rajdas98: not all clusters, if he selects ignite-cluster. he will show the internal cluster.

rajdas98: Made the cluster_type not mandatory

PR #1866, opened 2020-08-14
(feat auth): usermanagement functionality 

Sanjay1611:
## Proposed changes

This PR consists of following changes 

- updates password

- logout functionality

- resets password

## Types of changes

## Checklist

## Dependency


litmus-portal/backend/auth/pkg/manage/manager.go
```diff
@@ -72,20 +86,40 @@ func (m *Manager) VerifyUserPassword(username, password string) (*models.PublicU
...
+	storedUser, err := m.GetUser(username)
+	if err != nil {
+		return err
+	}
+	storedUser.LoggedIn = true
+	return m.userStore.UpdateUser(storedUser)
+}
+
+// LogoutUser verifies user password
+func (m *Manager) LogoutUser(username string) error {
```
gdsoumya, line 100: This logic won't work, if the user closes the browser directly without logging out. This would not be triggered. If we are planning to track the logins this method is not optimal.

litmus-portal/backend/auth/pkg/models/user.go
```diff
@@ -141,6 +149,36 @@ func (uinfo *PublicUserInfo) GetEmail() string {
 	return uinfo.Email
 }
 
+// GetCreatedAt user createdAt
+func (uinfo *PublicUserInfo) GetCreatedAt() *time.Time {
```
gdsoumya, line 153: are these getter functions required? I can see the variables are exported so they can be directly accessed. Getter functions are used for private or inaccessible variables mostly.

litmus-portal/backend/auth/pkg/models/user.go
```diff
@@ -58,71 +65,72 @@ const (
 )
 
 // GetID user id
-func (u *User) GetID() string {
+func (u *UserCredentials) GetID() string {
```
gdsoumya, line 68: Same for these getter functions, the variables are already exported

litmus-portal/backend/auth/pkg/manage/manager.go
Sanjay1611: We can extend this  logout functionality with the help of jwt. But we need to have this endpoint in order to update the database about the login info. We can surely extend it in the next releases

litmus-portal/backend/auth/pkg/models/user.go
Sanjay1611: Having getter functions is a good practice in golang. You can see that in Kubernetes go API itself. ALl its variables are exported but still, we do have getter functions for them

PR #1878, opened 2020-08-16
chore(litmus-portal): Add unified modal component

saswatamcode:
## Proposed changes
Making a unified modal component, `Unimodal`. This unified modal does not need to be styled again when it needs to be used. Only content passed inside the modal needs to be styled. Can pass flags such as `hasCloseBtn` to automatically add a close button to the top right and `isDark` to make modal background dark. Most modals have been adapted to use the unified modal. Pertaining to issue #1887 
## Types of changes

## Checklist

This does not bring about a difference in the modal responsiveness behaviour as such. That needs to be handled separately.

### DIFF
--- a/litmus-portal/frontend/src/components/Unimodal/index.tsx
+++ b/litmus-portal/frontend/src/components/Unimodal/index.tsx
@@ -6,11 +6,12 @@ import useStyles from './styles';
 /* DelUser, NewUserModal, ResetModal need to be shifted */
 
 interface UnimodalProps {
-  children?: any;
+  children?: React.ReactNode;
   isOpen: boolean;
-  handleClose: any;
+  handleClose: () => void;
   hasCloseBtn: boolean;
   isDark?: boolean;
+  textAlign?: string;
 }
 
 const Unimodal: React.FC<UnimodalProps> = ({
@@ -19,8 +20,10 @@ const Unimodal: React.FC<UnimodalProps> = ({
   handleClose,
   hasCloseBtn,
   isDark,
+  textAlign,
 }) => {
-  const classes = useStyles();
+  const styleProps = { textAlign, isDark };
+  const classes = useStyles(styleProps);
 
   return (
     <Modal
@@ -32,15 +35,13 @@ const Unimodal: React.FC<UnimodalProps> = ({
       aria-labelledby="simple-modal-title"
       aria-describedby="simple-modal-description"
     >
-      <div
-        className={isDark ? classes.darkModalContainer : classes.modalContainer}
-      >
+      <div className={classes.modalContainer}>
         {hasCloseBtn && (
           <div className={classes.modalContainerClose}>
             <Button
               variant="outlined"
               color="secondary"
-              className={isDark ? classes.darkCloseButton : classes.closeButton}
+              className={classes.closeButton}
               onClick={handleClose}
             >
               &#x2715;
--- a/litmus-portal/frontend/src/components/Unimodal/index.tsx
+++ b/litmus-portal/frontend/src/components/Unimodal/index.tsx
@@ -1,0 +1,49 @@
+import React from 'react';
+import Modal from '@material-ui/core/Modal';
+import Button from '@material-ui/core/Button';
+import useStyles from './styles';
+
+/* WelcomeModal, PasswordModal, FinishModal, PersonalDetails, ChooseWorkflow, ReliabiltyScore, VerifyCommit */
+
+interface UnimodalProps {
+  children?: any;
+  isOpen: boolean;
+  handleClose: any;
+  hasCloseBtn: boolean;
+}
+
+const Unimodal: React.FC<UnimodalProps> = ({
+  children,
+  isOpen,
+  handleClose,
+  hasCloseBtn,
+}) => {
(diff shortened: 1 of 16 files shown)


ishangupta-ds:
@saswatamcode you need to raise a PR to the master branch i guess, the base branch has been changed. cc: @rajdas98

litmus-portal/frontend/src/components/WelcomeModal/styles.ts
```diff
@@ -3,13 +3,7 @@ import { makeStyles, Theme } from '@material-ui/core/styles';
...
   rootContainer: {
-    margin: '3rem auto',
-    width: '60%',
-    height: '85%',
-    background: theme.palette.common.white,
-    borderRadius: 3,
-    textAlign: 'center',
-    outline: 'none',
+    paddingBottom: '8.5rem',
   },
```
Saranya-jena, line 7: Would be a lot better if we can use `theme.spacing()` for padding and margins.

litmus-portal/frontend/src/components/Unimodal/index.tsx
```diff
@@ -0,0 +1,56 @@
+import React from 'react';
+import Modal from '@material-ui/core/Modal';
+import Button from '@material-ui/core/Button';
+import useStyles from './styles';
+
+/* DelUser, NewUserModal, ResetModal need to be shifted */
+
+interface UnimodalProps {
+  children?: any;
```
S-ayanide, line 9: Don't keep the type as any, instead replace it with a `React.ReactNode` since the children component would be a potential DOM node

litmus-portal/frontend/src/components/Unimodal/index.tsx
```diff
@@ -0,0 +1,56 @@
...
+  isOpen: boolean;
+  handleClose: any;
```
S-ayanide, line 11: handleClose would be a function not `any` you can take a reference from `InputField` component

litmus-portal/frontend/src/components/Unimodal/styles.ts
```diff
@@ -0,0 +1,58 @@
+import { makeStyles } from '@material-ui/core';
+
+const useStyles = makeStyles((theme) => ({
+  modalContainer: {
+    width: '70%',
+    padding: '1rem',
+    margin: '2rem auto',
+    background: theme.palette.common.white,
+    borderRadius: 3,
+    textAlign: 'center',
```
S-ayanide, line 10: This is a good change since we have text centred in most of the modals but if you take a look at the User management modal the texts are aligned to the left. You might either want to pass a prop of textAlign to the Unimodal component and set this property here or just completely remove it and expect the developers to wrap the in the styles on the child itself.

>The former one is preferred

litmus-portal/frontend/src/components/Unimodal/styles.ts
```diff
@@ -0,0 +1,58 @@
+import { makeStyles } from '@material-ui/core';
+
+const useStyles = makeStyles((theme) => ({
+  modalContainer: {
+    width: '70%',
```
S-ayanide, line 5: Also fix the height to 80 or 70% what seems best to you and make the content inside vertically scrollable. That would abstract out most of the reusability of the modal.
```
overflow-x: hidden;
overflow-y: auto;
```

litmus-portal/frontend/src/components/Unimodal/styles.ts
```diff
@@ -0,0 +1,58 @@
...
+    borderRadius: 3,
+    textAlign: 'center',
+    outline: 'none',
+  },
+
+  darkModalContainer: {
+    width: '70%',
+    padding: '1rem',
+    margin: '2rem auto',
+    background: '#161616',
```
S-ayanide, line 18: Use this from the theme

litmus-portal/frontend/src/components/Unimodal/styles.ts
```diff
@@ -0,0 +1,58 @@
...
+    textAlign: 'center',
```
S-ayanide, line 19: Again, do it the other way

litmus-portal/frontend/src/components/WelcomeModal/styles.ts
saswatamcode: Okay. I'll change this to `theme.spacing()`. There's a mix of `theme.spacing()` and `rem` used in the styling of separate components which will probably be made consistent later.

litmus-portal/frontend/src/components/Unimodal/index.tsx
saswatamcode: I was actually confused about which one to use, `JSX.Element` or `ReactNode`. So I followed the advice [here](https://github.com/Microsoft/TypeScript/issues/6471). Will change this to `ReactNode`.

litmus-portal/frontend/src/components/Unimodal/styles.ts
saswatamcode: I actually used the hex code from the `YamlEditor` component as the `VerifyCommit` component showed that inside a modal. Will change this to use `theme.palette.common.black`.

PR #1941, opened 2020-08-31
Updated Self-Deployment and Subscriber Manifest

### LINKED ISSUE #1940: Integrate self-deployer with admin project creation in GQL server
- Self-Deployer should only start after admin creates their project
- Need to update the manifests accordingly
- Perform self-cluster registration on GQL-Server

gdsoumya:

## Proposed changes

- Self-Deployer starts only after admin project is created
- Updated the subscriber manifest, now all the components can be installed in one-go

Fixes #1940 

## Types of changes

## Checklist


rajdas98:
In this PR, we don't need to use a dynamic client or self-deployer manifest. As we are known to resource
Here we can use simple client-go and using its functions to create a self-deployer deployment

litmus-portal/backend/workflow-agent/pkg/cluster/events/util.go
```diff
@@ -1,13 +1,14 @@
 package events
 
 import (
+	v1 "Projects/go/pkg/mod/k8s.io/apimachinery@v0.17.9/pkg/apis/meta/v1"
```
rajdas98, line 4: Why are we using the local pkg?

gdsoumya: Not sure when that changed. Fixed it.

gdsoumya:
> In this PR, we don't need to use a dynamic client or self-deployer manifest. As we are known to that resource i.e., `Deployment`
> Here we can use simple client-go and using its functions to create a self-deployer deployment with its ENV

We can do that, but I believe having it read the data from the yaml is much more readable and easy to maintain as we can just make changes to the yaml directly instead of the code. Also Building the docker image will also be easier if we just make changes to the yaml as everything else will be taken from cache.

rajdas98:
It's a single single deployment, not required to be in separate. If you use simple client go, it will reduce so many extra steps and create deployment object is also readable

gdsoumya:
> It's a single single deployment, not required to be in separate file. If you use simple client go, it will reduce so many extra steps and creating a k8s deployment object is also readable

We will have to update the code directly everytime we change something in the manifest, it might reduce a few steps to do it directly but currently we just have to make 1 function call to apply it.

rajdas98 (approved):

PR #1958, opened 2020-09-03
Added Error field in Payload and Subscriber restart on connection failure

### LINKED ISSUE #1957: Subscriber doesn't reconnect if GQL-Server restarts
**What happened**:

Subscriber doesn't reconnect with the gql-server if the server restarts also the cluster-connect subscription doesn't handle the errors sent by the server.

- Running the subscriber with wrong cluster credentials won't throw an error
- After connecting subscriber to gql-server if the server is killed the subscriber will not reconnect.

gdsoumya:

## Proposed changes

- Added Error check for graphql cluster-connect errors
- Restart Subscriber on Connection Failure

Fixes #1957 

## Types of changes

## Checklist


litmus-portal/backend/subscriber/pkg/gql/subscriptions.go
```diff
@@ -80,7 +83,7 @@ func ClusterConnect(clusterData map[string]string) {
 			// send pod logs
 			logrus.Print("LOG REQUEST ", podRequest)
 			SendPodLogs(clusterData, podRequest)
-		} else {
+		} else if strings.ToLower(r.Payload.Data.ClusterConnect.Action.RequestType) == "create" {
```
rajdas98, line 86: RequestType can be `update`, `delete`, `get`. We should not put one type, this type will be decided by clusteroperation function

litmus-portal/backend/subscriber/pkg/gql/subscriptions.go
```diff
@@ -55,8 +55,7 @@ func ClusterConnect(clusterData map[string]string) {
 	for {
 		_, message, err := c.ReadMessage()
 		if err != nil {
-			log.Println("read:", err)
-			return
+			log.Fatal("SUBSCRIPTION ERROR : ", err)
```
rajdas98, line 58: Why we are removing return

rajdas98 (changes requested):

litmus-portal/backend/subscriber/pkg/gql/subscriptions.go
gdsoumya: The subscriber needs to halt and restart, to reconnect to the subscription.

litmus-portal/backend/subscriber/pkg/gql/subscriptions.go
gdsoumya: This needs a change in the workflow query because I kept the requesttype as the type of action that's needed. I can make it like "update" || "create" || "delete"... That should also solve it.

rajdas98: We don't have to make another check, it's already there in clusteroperations

gdsoumya: Updated it, now it should work for all the 4 types.

rajdas98 (approved):

PR #2085, opened 2020-09-16
(litmus-portal): Fix the styling and css issues

chandankumar4:
Signed-off-by: Chandan Kumar <chandan.kumar@mayadata.io>

## Proposed changes

- This commit will fix the CSS styling issue and make it configurable to use theme.
- Some of the extra variables are added which are an exceptional case.

## Types of changes

## Checklist

### DIFF
--- a/litmus-portal/frontend/public/index.css
+++ b/litmus-portal/frontend/public/index.css
@@ -1,4 +1,3 @@
 body > #root > div {
   height: 100vh;
-  background-color: white;
 }
--- a/litmus-portal/frontend/public/index.css
+++ b/litmus-portal/frontend/public/index.css
@@ -1,3 +1,4 @@
 body > #root > div {
   height: 100vh;
+  background-color: white;
 }
--- a/litmus-portal/frontend/src/components/Header/styles.ts
+++ b/litmus-portal/frontend/src/components/Header/styles.ts
@@ -5,14 +5,13 @@ const useStyles = makeStyles((theme: Theme) => ({
   appBar: {
     width: '100%',
     height: '100%',
-    backgroundColor: theme.palette.background.default,
+    background: theme.palette.background.default,
   },
 
   headerFlex: {
     height: '5%',
     flexDirection: 'row',
     justifyContent: 'space-between',
-    backgroundColor: theme.palette.background.default,
     marginTop: theme.spacing(1),
     marginBottom: theme.spacing(0.5),
     paddingTop: theme.spacing(1),
@@ -23,7 +22,6 @@ const useStyles = makeStyles((theme: Theme) => ({
     height: '5%',
     flexDirection: 'row',
     justifyContent: 'space-between',
-    backgroundColor: theme.palette.background.default,
     marginTop: theme.spacing(0),
     marginBottom: theme.spacing(0),
     paddingTop: theme.spacing(0),
@@ -36,7 +34,6 @@ const useStyles = makeStyles((theme: Theme) => ({
     height: '5%',
     flexDirection: 'row',
     justifyContent: 'space-between',
-    backgroundColor: theme.palette.background.default,
     marginTop: theme.spacing(0),
     marginBottom: theme.spacing(0),
     paddingTop: theme.spacing(1),
@@ -47,7 +44,6 @@ const useStyles = makeStyles((theme: Theme) => ({
     height: '5%',
     flexDirection: 'row',
     justifyContent: 'space-between',
-    backgroundColor: theme.palette.background.default,
     marginTop: theme.spacing(0.5),
     marginBottom: theme.spacing(0),
     paddingTop: theme.spacing(2),
@@ -63,7 +59,7 @@ const useStyles = makeStyles((theme: Theme) => ({
   projectDisplay: {
     marginTop: theme.spacing(-1.25),
     fontSize: 12,
-    color: 'grey',
+    color: theme.palette.projectDisplayColor,
   },
   // Style for ProfileDropdownSection and ProfileDropdownItems.
   avatarBackground: {
@@ -155,6 +151,7 @@ const useStyles = makeStyles((theme: Theme) => ({
     paddingBottom: theme.spacing(1),
     paddingLeft: theme.spacing(8),
     paddingRight: theme.spacing(8),
+    color: theme.palette.buttonEditProfile,
   },
 
   dividerTop: {
--- a/litmus-portal/frontend/src/components/InputField/styles.ts
+++ b/litmus-portal/frontend/src/components/InputField/styles.ts
@@ -9,6 +9,7 @@ const useStylesLitmus = makeStyles((theme: Theme) =>
           : `1px solid ${theme.palette.primary.dark}`,
       overflow: 'hidden',
       borderRadius: 4,
+      color: 'inherit',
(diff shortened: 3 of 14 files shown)


litmus-portal/frontend/src/theme/index.tsx
```diff
@@ -136,6 +150,13 @@ function customTheme(options: ThemeOptions) {
         passedTests: 'rgba(16, 155, 103, 0.2)',
         failedTests: 'rgba(202, 44, 44, 0.2)',
       },
+      projectDisplayColor: 'rgba(128, 128, 128)',
+      buttonEditProfile: 'rgba(0, 0, 0)',
+      homePageCardBackgroundColor: '#FFFFFF',
+      homePageWorkflowCardBackgroundColor: '#FFFFFF',
+      workflowTitleColor: '#5B44BA',
```
arkajyotiMukherjee, line 157: this one is same as secondary.dark...in the other projects do we not have this as secondary.dark? If not we can define it like this

litmus-portal/frontend/src/theme/index.tsx
```diff
@@ -136,6 +150,13 @@ function customTheme(options: ThemeOptions) {
...
+      layoutBackground: '#FAFBFD',
+      sidebarBackground: '#FAFBFD',
```
arkajyotiMukherjee, line 159: these two are same as background.default...can we use that?

litmus-portal/frontend/src/theme/index.tsx
```diff
@@ -136,6 +150,13 @@ function customTheme(options: ThemeOptions) {
         passedTests: 'rgba(16, 155, 103, 0.2)',
         failedTests: 'rgba(202, 44, 44, 0.2)',
       },
+      projectDisplayColor: 'rgba(128, 128, 128)',
+      buttonEditProfile: 'rgba(0, 0, 0)',
+      homePageCardBackgroundColor: '#FFFFFF',
```
arkajyotiMukherjee, line 155: can we have a common `card.background: #fff`

arkajyotiMukherjee (commented):
These are not absolute changes...just suggestions

litmus-portal/frontend/src/theme/index.tsx
chandankumar4: Yes, these are the exceptional case, in which we are not using the secondary color in other design

litmus-portal/frontend/src/theme/index.tsx
chandankumar4: No, we can not use the common `card.background` because as per design some card have different background color! Once we fix the design then we can remove this exceptional variable but a/c to current design these all are required.

arkajyotiMukherjee (approved):
LGTM

arkajyotiMukherjee (dismissed):
LGTM

ishangupta-ds (dismissed):
LGTM.

litmus-portal/frontend/public/index.css
```diff
@@ -1,3 +1,4 @@
 body > #root > div {
   height: 100vh;
+  background-color: white;
```
S-ayanide, line 3: Is this background color change in /public necessary?

litmus-portal/frontend/src/components/Header/styles.ts
```diff
@@ -63,7 +59,7 @@ const useStyles = makeStyles((theme: Theme) => ({
   projectDisplay: {
     marginTop: theme.spacing(-1.25),
     fontSize: 12,
-    color: 'grey',
+    color: theme.palette.projectDisplayColor,
```
S-ayanide, line 62: Can we have a better name for the themes! Like `theme.palette.project.grey`

litmus-portal/frontend/src/components/InputField/styles.ts
```diff
@@ -78,7 +80,7 @@ const useStyles = makeStyles((theme) => ({
     borderColor: theme.palette.error.main,
   },
   inputArea: {
-    width: theme.spacing(45),
+    width: 'inherit',
```
S-ayanide, line 83: Inheriting the width will make the input fields go out of shape for Login, Settings and other screens where there were fixed width before. If we face a use case where we want bigger input fields we can pass the width as a style `style:{{width: <size>}}`

litmus-portal/frontend/src/containers/app/App-styles.ts
```diff
@@ -2,7 +2,7 @@ import { makeStyles, Theme } from '@material-ui/core/styles';
 
 const useStyles = makeStyles((theme: Theme) => ({
   root: {
-    width: '100vw',
+    width: 'auto',
```
S-ayanide, line 5: If you keep auto width on root in case any of the components or layout takes up extra width the main frame would go out of range and give a horizontal scrollbar. Which is constrained with 100vw.

S-ayanide (changes requested):

litmus-portal/frontend/src/components/Header/styles.ts
chandankumar4: color name we can't put in the variable name because it will change based on the project

litmus-portal/frontend/src/components/InputField/styles.ts
chandankumar4: I have used `inherit` here because there is a requirement of width change and it's defined in the component itself. So now when we use this component we can pass the width from the parent. but a/c to your approach can we change the width without changing in the components?

litmus-portal/frontend/src/containers/app/App-styles.ts
chandankumar4: I was facing the issue with `width: 100vw` the `horizontal` scroll was coming

litmus-portal/frontend/public/index.css
chandankumar4: It's necessary here because by default it will pick the `white` as background but in another project we need some different background color, but anyway it's not coming from the theme so we can remove it

litmus-portal/frontend/src/components/Header/styles.ts
S-ayanide: No I mean if we could just change the name from projectDisplayColor to project Object and have grey assigned to any hex value that you are using in your project.

Like
```
project: {
        grey: '#value',
},
```

litmus-portal/frontend/src/components/InputField/styles.ts
S-ayanide: App width is always fixed and it y default takes the entire screen size, we can define the width of any components we are using on top of it. By default it is `100%` I believe, by making it auto we are opening the possibility for it to change. I guess you can use width auto on any component that you are passing from the parent not on App itself.

S-ayanide (approved):

litmus-portal/frontend/src/components/Header/styles.ts
chandankumar4: yes we can do that, but here grey is denoting a color, so let say we need the green color so the value will be `gray: green`, which will create confusion

S-ayanide (approved):

arkajyotiMukherjee (approved):
LGTM

PR #2105, opened 2020-09-23
GQL Mutation for getWorkflows (Workflows + Workflowruns)

rajdas98:
## Proposed changes

- GQL Mutation for getWorkflows (Workflows + Workflowruns)
- Fixing corrupted gqlgen 
## Types of changes

## Checklist

## Dependency


ishangupta-ds (approved):

litmus-portal/graphql-server/graph/resolver.go
```diff
@@ -28,4 +28,4 @@ func NewConfig() generated.Config {
 		return next(ctx)
 	}
 	return config
-}
+}
```
gdsoumya, line 31: can you do gofmt once

gdsoumya (changes requested):

litmus-portal/graphql-server/graph/schema.graphqls
```diff
@@ -173,7 +197,10 @@ type Query{
 
   users: [User!]! @authorized
 
+  # [Deprecated soon]
   getScheduledWorkflows(project_id: String!): [ScheduledWorkflows]! @authorized
+
+  getWorkflows(project_id: String!): [Workflows]! @authorized
```
gdsoumya, line 203: I think @ishangupta-ds needed to select only a few workflows and not all workflows, can we add another parameter - an array of string that will have the workflow-ids. If it is empty we send all else only those specific workflows

gdsoumya (dismissed):

litmus-portal/graphql-server/graph/schema.graphqls
rajdas98: Purpose of the API is to deprecate `getScheduledWorkflows` and `getWorkFlowRuns`. and if he wants workflows by ID. we can make another mutation getWorkflowByID

rajdas98: There will be two mutation for workflow (List-->getWorkflows, Get --> getWorkflowByID )

rajdas98: We can change the naming convention to getWorkflow to Listworkflow

gdsoumya: We can have the same mutation do both right? that reduces extra queries, the change will not be too much, the array could be kept optional. That way the frontend can only send the projectID and get all the workflows or send an array also to get specific workflows.

rajdas98: Yes, But every API should have a single purpose. I have added a new Query to get workflow by ID

rajdas98: Just like CRUDL

gdsoumya: Can you change it to an array of ids, if we want more than 1 we have to perform multiple queries

rajdas98: If they want an array of workflow ids they will call `ListWorkflow`. If single workflow, call `GetWorkflow`

gdsoumya: I meant array of specific workflows, ListWorkflows will give all the data. I think that was the main problem that @ishangupta-ds had.

ishangupta-ds: yeah @rajdas98 ... @gdsoumya is right about the requirement, we need the ID selector too, this API too solves the issue but the thing is we can have a single API if user gives an array of ID send data for only those workflow schedules i.e. workflow ids and if the array is empty send all workflow data in mongoDB. If you planning to make 2 APIs then okay but if just one then it requires a ID filter which will be an array to filter out just the data of workflows that we need for frontend.

ishangupta-ds (changes requested):
can you make this API a generic one with workflow ids filter ... if array of ids is there send just those if array is empty send all. @rajdas98 cc: @gdsoumya  Making 2 separate APIs is not necessary as per data requirement. one query can work for both with this minor tweak.

oumkale (approved):

litmus-portal/graphql-server/graph/schema.graphqls
```diff
@@ -173,7 +197,12 @@ type Query{
 
   users: [User!]! @authorized
 
+  # [Deprecated soon]
   getScheduledWorkflows(project_id: String!): [ScheduledWorkflows]! @authorized
+
+  ListWorkflow(project_id: String!, workflow_ids: [ID] ): [Workflow]! @authorized
+
+  GetWorkflow(workflow_id: String!): Workflow @authorized
```
gdsoumya, line 205: We don't need this anymore right? ListWorkflows should handle this use-case?

litmus-portal/graphql-server/graph/schema.resolvers.go
```diff
@@ -104,6 +104,18 @@ func (r *queryResolver) GetScheduledWorkflows(ctx context.Context, projectID str
 	return queries.QueryWorkflows(projectID)
 }
 
+func (r *queryResolver) ListWorkflow(ctx context.Context, projectID string, workflowIds []string) ([]*model.Workflow, error) {
+	if len(workflowIds) == 0 {
```
gdsoumya, line 108: Will it be empty array or `nil`  when the frontend doesn't send a parameter at all?

litmus-portal/graphql-server/pkg/database/mongodb/operations.go
```diff
@@ -90,6 +107,24 @@ func GetWorkflowsByProjectID(project_id string) ([]ChaosWorkFlowInput, error) {
 	return workflows, nil
 }
 
+func GetWorkflowByID(workflow_id string) (*ChaosWorkFlowInput, error) {
```
gdsoumya, line 110: We can use the `GetWorkflowsByIDs` for this function too

gdsoumya (changes requested):

litmus-portal/graphql-server/graph/schema.graphqls
rajdas98: Yes, but, In ListWorkflow we have to send `project_id`. GetWorkflow is like Get operation for Workflows

litmus-portal/graphql-server/graph/schema.resolvers.go
rajdas98: It will work if workflowIDs is not present. Graphql convert it to nil

litmus-portal/graphql-server/graph/schema.graphqls
gdsoumya: We actually need to keep the project_id in this one too, it is required for authorization. We can just verify if the user is part of the project and then proceed with getting the workflow.

rajdas98: Okay, then this can be removed.

ishangupta-ds (approved):
LGTM.

gdsoumya (approved):

PR #2162, opened 2020-10-01
Update README.md

mofolactic:
I have suggested some better words. This will make for the reader easy to understand.


arkajyotiMukherjee (changes requested):
Can you also change `Guildelines` --> `Guidelines` in line no 46
Thank you for showing interest in Litmuschoas. do star our repo if you want to follow future updates

PR #2226, opened 2020-10-09
Adding endpoint to ClusterReg mutation response

rajdas98:

## Proposed changes

Adding endpoint to ClusterReg mutation response

## Types of changes

## Checklist

## Dependency

### DIFF
--- a/litmus-portal/graphql-server/pkg/graphql/mutations/mutation.go
+++ b/litmus-portal/graphql-server/pkg/graphql/mutations/mutation.go
@@ -2,6 +2,7 @@ package mutations
 
 import (
 	"encoding/json"
+	"github.com/litmuschaos/litmus/litmus-portal/graphql-server/pkg/k8s"
 	"log"
 	"strconv"
 	"strings"
@@ -30,6 +31,11 @@ func ClusterRegister(input model.ClusterInput) (*model.ClusterRegResponse, error
 		return &model.ClusterRegResponse{}, err
 	}
 
+	endPoint, err := k8s.GetPortalEndpoint()
+	if err != nil {
+		return nil, err
+	}
+
 	newCluster := database.Cluster{
 		ClusterID:    clusterID,
 		ClusterName:  input.ClusterName,
@@ -41,6 +47,7 @@ func ClusterRegister(input model.ClusterInput) (*model.ClusterRegResponse, error
 		CreatedAt:    strconv.FormatInt(time.Now().Unix(), 10),
 		UpdatedAt:    strconv.FormatInt(time.Now().Unix(), 10),
 		Token:        token,
+		Endpoint:     endPoint,
 	}
 
 	err = database.InsertCluster(newCluster)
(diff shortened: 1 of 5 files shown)


oumkale:
Please will you add the same in Query as well?

oumkale (approved):

litmus-portal/graphql-server/pkg/graphql/mutations/mutation.go
```diff
@@ -54,6 +61,7 @@ func ClusterRegister(input model.ClusterInput) (*model.ClusterRegResponse, error
 		ClusterID:   newCluster.ClusterID,
 		Token:       token,
 		ClusterName: newCluster.ClusterName,
+		Endpoint:    endPoint,
```
gdsoumya, line 64: There is no need to persist it in the DB as the endpoint will be the same everytime?

gdsoumya (changes requested):

litmus-portal/graphql-server/pkg/graphql/mutations/mutation.go
rajdas98: No endpoint will not same every time. It depends which node the pod had been schedule.

gdsoumya: Then the whole external cluster thing will fail, if the ip keeps changing how will the subscriber connect?

rajdas98: Also, it required to save in db because after logout frontend will lose the state

rajdas98: Yes this is a existing problem. Pod can be schedule to another node.if the node has some problem

gdsoumya: Backend can always send the same endpoint, so state management is not required. But this is a bigger problem if the pod is rescheduled all external clusters are disconnected? We can't use this node ip approach then

rajdas98: This can be solved by ingress , for now it's okay. Where will frontend stores the endpoint if we logout?

rajdas98: It is required for the getcluster query, when user wants to see the pending cluster endpoint

gdsoumya: If the endpoint is constant we can just retrieve that endpoint and always provide frontend with that, so no need to store. We should use ingress rather than nodeIP then. This will cause problems.

gdsoumya: And also we should be sending the updated endpoint and not the old nodeIP which could've changed. Storing in db will give old endpoint which will be useless.

rajdas98: Ingress is one approach, other approach is load balancer. In ingress, user will need a domain to setup. It's not easy to setup.

gdsoumya: Even if we don't want to setup ingress or loadbalancer right now, storing nodeIP in db will not help as I already stated above.

PR #2290, opened 2020-10-22
Minor fix for gauge count errors on sock-shop dashboard.

ishangupta-ds:
Signed-off-by: ishangupta-ds <ishan.gupta@mayadata.io>

## Proposed changes

Solves issue for decimal gauge values.

When Instant if off it uses the query API endpoint rather than the query_range API endpoint on Prometheus, which is more efficient if we only care about the end of the time range and don't want to pull in data that Grafana is going to give away again. Making instant as true on the gauge solves the issue of decimal numbers for the count.

## Types of changes

## Checklist

## Dependency
- None.


ishangupta-ds:
Screenshot from 2020-10-21 16-29-54

Screenshot 2020-10-22 at 4 04 01 PM

ksatchit (approved):

PR #2331, opened 2020-11-09
chore(litmus-portal): Patch to keep the workflow runs when schedules are deleted

amityt:
## Proposed changes

Adds a field `isRemoved` which is set to false when a Chaos Workflow is scheduled.
Once a Schedule is deleted , `isRemoved` is set to true.

1.  **Data when Chaos Workflow is Scheduled**
Screenshot from 2020-11-09 09-44-19

2.  **Data when Chaos Schedule is deleted**
Screenshot from 2020-11-09 09-47-40

## Types of changes

## Checklist

## Dependency


litmus-portal/graphql-server/pkg/database/mongodb/operations.go
```diff
@@ -166,15 +166,14 @@ func InsertChaosWorkflow(chaosWorkflow ChaosWorkFlowInput) error {
 
 func DeleteChaosWorkflow(workflowid string) (bool, error) {
```
rajdas98, line 167: Remove `DeleteChaosWorkflow` and call `UpdateChaosWorkflow` with these params

rajdas98 (changes requested):

litmus-portal/graphql-server/pkg/database/mongodb/operations.go
```diff
@@ -166,15 +166,14 @@ func InsertChaosWorkflow(chaosWorkflow ChaosWorkFlowInput) error {
...
 	ctx, _ := context.WithTimeout(backgroundContext, 10*time.Second)
-	res, err := workflowCollection.DeleteOne(ctx, bson.M{"workflow_id": workflowid})
+	update := bson.M{"$set": bson.M{"isRemoved": true}}
```
gdsoumya, line 169: The workflow needs to be deleted from the cluster also, or else it will keep generating the new workflow runs in case of scheduled workflow

gdsoumya (changes requested):

rajdas98:
Also, Change get workflow function. It should list down based on `isRemoved: false`

litmus-portal/graphql-server/pkg/database/mongodb/operations.go
gdsoumya: cc: @rajdas98

amityt: @gdsoumya We are deleting the manifest too, will that not work?

rajdas98: I think for this we have to change the logic in the upsert function. Its entry is not present don't insert it.

rajdas98: func UpdateWorkflowRun

gdsoumya: No I meant deleting from the cluster where the workflow is running, just deleting from db doesn't stop the workflow from running.

rajdas98: Deleting from cluster functionality is present

rajdas98 (approved):

gdsoumya (approved):

PR #2332, opened 2020-11-10
Replacing agent namespace with workflow manifest namespace

rajdas98:

## Proposed changes

Replacing agent namespace with workflow manifest namespace

## Types of changes

## Checklist

## Dependency

### DIFF
--- a/litmus-portal/graphql-server/pkg/graphql/mutations/mutation.go
+++ b/litmus-portal/graphql-server/pkg/graphql/mutations/mutation.go
@@ -3,6 +3,7 @@ package mutations
 import (
 	"encoding/json"
 	"log"
+	"os"
 	"strconv"
 	"strings"
 	"time"
@@ -19,6 +20,7 @@ import (
 	"github.com/litmuschaos/litmus/litmus-portal/graphql-server/pkg/graphql/subscriptions"
 	"github.com/litmuschaos/litmus/litmus-portal/graphql-server/utils"
 	"github.com/pkg/errors"
+	"github.com/tidwall/gjson"
 	"github.com/tidwall/sjson"
 	"go.mongodb.org/mongo-driver/bson"
 )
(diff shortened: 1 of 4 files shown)


litmus-portal/graphql-server/pkg/graphql/mutations/mutation.go
```diff
@@ -307,11 +317,18 @@ func UpdateWorkflow(workflow *model.ChaosWorkFlowInput, r store.StateData) (*mod
 		return nil, err
 	}
 
+	workflowNamespace := gjson.Get(workflow.WorkflowManifest, "metadata.namespace").String()
```
gdsoumya, line 320: Is there a need to add a new package? json.Unmarshal would work too?

rajdas98: WorkflowManifest is JSON string and to unMarshal WorkflowManifest we need a struct of workflow manifest

gdsoumya: We can use map[string]inteface{} to locate the "namespace" key, if it is there we use it or else we use the os.Getenv()

rajdas98: Yes, we can use but just to ease the getting of the namespace. I have used it. 
We have used similar pkg for setting also

rajdas98: FYI map[string]map[string]interface{} or map[string]interface{} is not able to parse the manifest

gdsoumya: Why is map[string]interface{} not able to parse it? I remember parsing it while I made the changes for the wf_id in cron-workflows?

gdsoumya: We can use this package, i think Ishan already used the sjson package to add the wf_id.

rajdas98: You have use sjson to set wf_id in cron-workflow, For parsing this kind of JSON, it requires proper struct.

gdsoumya: No I did not use sjson, it was already there I would've used simple map for that also. 

This is the unmarshaling code I used 
```
var workflow map[string]interface{}
	err := json.Unmarshal([]byte(input.WorkflowManifest), &workflow)
	if err != nil {
		return nil, err
	}
```

gdsoumya: To get the ns you would have to marshal twice maybe, so it's easier to use this gjson package

rajdas98: Okay, Is it good to merge then?

gdsoumya (approved):

PR #2522, opened 2021-03-10
Updated mongo queries to use UID as unique identifier

SarthakJain26:
Signed-off-by: SarthakJain26 <sarthak.jain@mayadata.io>

## Proposed changes

- Updated MemberInput interface (replaced username with userID)
- Updated mongo queries to fetch details using UID

## Types of changes

## Checklist

## Dependency


rajdas98:
Now, we have moved to user_id. Is this function required
```
func GetUserByUserName(ctx context.Context, username string) (*User, error) {
	// ctx, _ := context.WithTimeout(backgroundContext, 10*time.Second)
	var user = new(User)
	query := bson.M{"username": username}
	err := userCollection.FindOne(ctx, query).Decode(user)
	if err != nil {
		log.Print("Error getting user with username: ", username, " error: ", err)
		return nil, err
	}
	return user, err
}

```

SarthakJain26:
> Now, we have moved to user_id. Is this function required
> 
> ```
> func GetUserByUserName(ctx context.Context, username string) (*User, error) {
> 	// ctx, _ := context.WithTimeout(backgroundContext, 10*time.Second)
> 	var user = new(User)
> 	query := bson.M{"username": username}
> 	err := userCollection.FindOne(ctx, query).Decode(user)
> 	if err != nil {
> 		log.Print("Error getting user with username: ", username, " error: ", err)
> 		return nil, err
> 	}
> 	return user, err
> }
> ```
@rajdas98 since the username is also a unique field, so while creating a new user we have to check if the username entered is not assigned to anyone else. We will need this function in such cases.

rajdas98 (approved):

PR #2580, opened 2021-03-21
Update README.md

rajdeepbhattacharya:
Added way to access minikube service url. Also added env variable LITMUS_PORTAL_NAMESPACE to NODE_NAME/EXTERNAL_IP/NODE_PORT variable.

## Proposed changes

## Types of changes

- [x] Documentation Update (if none of the other choices applies)

## Checklist

## Dependency

### DIFF
--- a/litmus-portal/README.md
+++ b/litmus-portal/README.md
@@ -61,11 +61,18 @@ kubectl apply -f https://raw.githubusercontent.com/litmuschaos/litmus/master/lit
 
 #### Retrieving external url to access the litmus portal
 
+##### GKE/Okteto Cloud/EKS
+
 ```bash
-export NODE_NAME=$(kubectl get pod -n litmus -l "component=litmusportal-frontend" -o=jsonpath='{.items[*].spec.nodeName}')
-export EXTERNAL_IP=$(kubectl get nodes $NODE_NAME -o jsonpath='{.status.addresses[?(@.type=="ExternalIP")].address}')
-export NODE_PORT=$(kubectl get -o jsonpath="{.spec.ports[0].nodePort}" services litmusportal-frontend-service -n litmus)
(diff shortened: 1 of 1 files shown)


ishangupta-ds (changes requested):
This env. variable is not used for cluster scoped installation @rajdeepbhattacharya in this case the command will fail to execute as intended, you may create separate steps for obtaining url for the namespaced install and cluster scoped install ( by default cluster manifest installs litmus in the litmus namespace ) cc: @rajdas98 @ajeshbaby

rajdeepbhattacharya:
> This env. variable is not used for cluster scoped installation @rajdeepbhattacharya in this case the command will fail to execute as intended, you may create separate steps for obtaining url for the namespaced install and cluster scoped install ( by default cluster manifest installs litmus in the litmus namespace ) cc: @rajdas98 @ajeshbaby

@ishangupta-ds actually in the docs env variable `LITMUS_PORTAL_NAMESPACE` is already taken as an input, then the whole setup should continue with the same namespace, not necessarily litmus namespace. I could see that in the yaml files also litmus namespace is used. The basis of my change request was only based on the user input and if you want to make the namespace dynamic based on user input. Also the node ip command doesn't work in minikube that's why I have shared the minikube command to get the endpoint ip.

rajdas98 (approved):
Looks good to me. 
@ishangupta-ds - It's just an env, I don't think it will break anything. Soon we will move to the helm as the main installation procedure.

rajdas98:
@rajdeepbhattacharya Can you update your PR branch?

ishangupta-ds:
@rajdas98 we may have to use a similar template for cluster scope installation where user has a choice to install portal in any namespace instead of the hardcoded litmus ns. We may need to add this change like @rajdeepbhattacharya suggests is on the docs, may be via helm we are giving the configuration option, the readme and cluster manifest are still having litmus as namespace.

ishangupta-ds (approved):

PR #2581, opened 2021-03-21
Using Quick action card from Litmus-ui and 'ops' to 'qps' change

whitetiger1399:
## Proposed changes
1. Replacing all custom QuickActionCard component imports with Litmus-UI quickActionCard 
2. All external Litmus docs will open in a new tab
3. `ops` in components/PreconfiguredDashboards/data.ts file to `qps`

## Types of changes

## Checklist

## Dependency


ishangupta-ds:
@whitetiger1399 since it’s being refactored can we import it from component folder and use in files, by switching the data based on a prop (feature/page) this can make it easier to manage as quick actions is a generic feature used in multiple pages. It can help in improving BCH score as well.

litmus-portal/frontend/src/views/MyHub/MyHubConnect/index.tsx
```diff
@@ -549,7 +549,7 @@ const MyHub = () => {
             {t('myhub.connectHubPage.videoDesc')}
           </Typography>
           <div className={classes.quickActionDiv}>
-            <QuickActionCard analyticsHome={false} nonAdmin />
+            <HomePageQuickActionCard />
```
ishangupta-ds, line 552: Can we use a single component named quick actions? from the generic components folder as it’s being reused at all places just with the change of data which may be switched using a prop depending on the page.

ishangupta-ds (changes requested):

litmus-portal/frontend/src/views/MyHub/MyHubConnect/index.tsx
whitetiger1399: Issue is not with the pros that is being passed to QuickActionCard but the way we want to pass the props. In Litmus, we can have a separate file with just data array specific to that page and then pass it to QuickActionCard. But for passing the text we require a functional component because we have to import usetransltion hook. For this every reason I took the quick action card from limtus-ui and then passed the props to it and made a local component specific to that page. If we keep the array in the same page file then the file will become big. Also for all pages having common QuickActionCard are using same local component. No redundant code.

ishangupta-ds: > In Litmus, we can have a separate file with just data array specific to that page and then pass it to QuickActionCard. But for passing the text we require a functional component because we have to import usetransltion hook. For this every reason I took the quick action card from limtus-ui and then passed the props to it and made a local component specific to that page. If we keep the array in the same page file then the file will become big. Also for all pages having common QuickActionCard are using same local component.

It’s not about redundant code, that way the code will be easier to manage as quick actions is not something feature related it’s generic thing which can hold context specific data, if the team approves then I am also fine with the current implementation too, though the suggested change would make it easy for community members to contribute.

ishangupta-ds:
The ops to qps is a very small change but it would have been better to keep it separate from this particular change as they are not related in any way, if you are adding this change please mention either in commit message or in PRs name, rather than just in description, it will be difficult to track it later. @whitetiger1399

ishangupta-ds (dismissed):
LGTM

ishangupta-ds (approved):

litmus-portal/frontend/src/views/MyHub/MyHubConnect/index.tsx
whitetiger1399: all requested changes implemented

Jonsy13 (approved):

amityt (approved):

vanshBhatia-A4k9 (approved):
Thanks for the variant based approach! :rocket:

ishangupta-ds (approved):

PR #2647, opened 2021-04-06
chore(v2): Alert and Validation added in workflow creation + Fixed minor issues

S-ayanide:
Signed-off-by: Sayan Mondal <sayan@chaosnative.com>

## Proposed changes

* Alerts added if the validation fails at any step of the workflow creation
* UI bug fixes
* Removed package-lock and added npmrc

## Types of changes

- [X] Bugfix (non-breaking change which fixes an issue)

## Checklist

- [X] I have read the [CONTRIBUTING](https://github.com/litmuschaos/litmus/blob/master/CONTRIBUTING.md) doc
- [X] I have signed the commit for DCO to be passed.
- [X] Lint and unit tests pass locally with my changes


litmus-portal/frontend/src/components/LitmusStepper/index.tsx
```diff
@@ -78,13 +77,17 @@ const LitmusStepper: React.FC<LitmusStepperProps> = ({
...
-            ? !hideNext && (
-                <ButtonFilled onClick={handleNext} disabled={disableNext}>
-                  <Typography>Next</Typography>
-                </ButtonFilled>
-              )
-            : finishAction}
+          {activeStep !== steps.length - 1 ? (
+            !hideNext && (
+              <ButtonFilled onClick={handleNext} disabled={disableNext}>
+                <Typography>Next</Typography>
```
ishangupta-ds, line 83: translation can be used.

litmus-portal/frontend/src/components/LitmusStepper/index.tsx
```diff
@@ -78,13 +77,17 @@ const LitmusStepper: React.FC<LitmusStepperProps> = ({
...
+              </ButtonFilled>
+            )
+          ) : (
+            <ButtonFilled onClick={handleNext} disabled={disableNext}>
+              <Typography>Finish</Typography>
```
ishangupta-ds, line 88: translation can be used.

litmus-portal/frontend/src/views/CreateWorkflow/TuneWorkflow/index.tsx
```diff
@@ -254,8 +256,11 @@ const TuneWorkflow = forwardRef((_, ref) => {
     }
   }, [engineDataLoading, experimentDataLoading]);
 
-  // console.log(generatedYAML);
   function onNext() {
+    if (customWorkflow && experiment.length === 0) {
```
ishangupta-ds, line 260: === 0 might not be needed !experiment.length can work

litmus-portal/frontend/src/views/CreateWorkflow/TuneWorkflow/styles.ts
```diff
@@ -6,6 +6,7 @@ const useStyles = makeStyles((theme) => ({
     color: theme.palette.text.primary,
     paddingTop: theme.spacing(2),
     margin: '0 auto',
+    marginBottom: theme.spacing(-2), // Overriding because LitmusStepper has default padding bottom
```
ishangupta-ds, line 9: can we override it in the stepper or provide a prop to do so, rather than this negative margin?

litmus-portal/frontend/src/views/CreateWorkflow/WorkflowSettings/index.tsx
```diff
@@ -146,7 +146,11 @@ const WorkflowSettings = forwardRef((_, ref) => {
       CRDLink,
     };
     localforage.setItem('workflow', workflowDetails);
-    if (validateWorkflowName(name)) {
+    if (name.length === 0) {
```
ishangupta-ds, line 149: !name.length can be used

litmus-portal/frontend/src/views/CreateWorkflow/WorkflowSettings/index.tsx
```diff
@@ -146,7 +146,11 @@ const WorkflowSettings = forwardRef((_, ref) => {
...
+      alert.changeAlertState(true); // Workflow Name is empty and user clicked on Next
+      return false;
+    }
+    if (name.length > 0 && validateWorkflowName(name)) {
```
ishangupta-ds, line 153: `name.length && validateWorkflowName(name)` can be used

ishangupta-ds (changes requested):
@S-ayanide PTAL

litmus-portal/frontend/src/views/CreateWorkflow/TuneWorkflow/styles.ts
S-ayanide: Actually, LitmusStepper uses a default padding top and bottom, changing that would affect all the stepper content and we only needed to override one single property for this screen only thus the negative margin.

ishangupta-ds (changes requested):
can we remove the package lock changes? it changes with every PR without any package changes. @S-ayanide cc: @rajdas98

litmus-portal/frontend/src/views/CreateWorkflow/TuneWorkflow/index.tsx
S-ayanide: It won't work without the assertion check because `if` condition works on a true or false basis but only `experiment.length` will return 0
Thus the condition becomes `if(true && 0){}`
Whereas here it states `if(true && true){}`

Just tested this locally

litmus-portal/frontend/src/views/CreateWorkflow/WorkflowSettings/index.tsx
S-ayanide: Same reason here as well it'll check for `if(0 or the length && true or false based on validation) -> if(number && boolean){}` which is not working thus the assertion is required.

litmus-portal/frontend/src/views/CreateWorkflow/WorkflowSettings/index.tsx
S-ayanide: > can we remove the package lock changes? it changes with every PR without any package changes. @S-ayanide cc: @rajdas98

I needed to install the new testing library and other dependencies that were newly added, npm install will update the package-lock, which is not in our control.

S-ayanide:
> can we remove the package lock changes? it changes with every PR without any package changes. @S-ayanide cc: @rajdas98

I needed to install the new testing library and other dependencies that were newly added, npm install will update the package-lock, which is not in our control.

litmus-portal/frontend/src/views/CreateWorkflow/TuneWorkflow/index.tsx
ishangupta-ds: @S-ayanide PTAL

> TCs:

Screenshot 2021-04-06 at 4 41 50 PM

> Results:

Screenshot 2021-04-06 at 4 42 10 PM

litmus-portal/frontend/src/views/CreateWorkflow/WorkflowSettings/index.tsx
ishangupta-ds: PTAL at the above one if its missing any edge case. If still it does not work we can go ahead with this version.

ishangupta-ds:
@S-ayanide the svc check is failing for frontend, build is skipped can you re run the checks or commit the necessary fix.

litmus-portal/frontend/src/views/CreateWorkflow/TuneWorkflow/index.tsx
arkajyotiMukherjee: !experiment.length will return true...but I feel since it's an if condition and multiple cases are there ===0 is more readable

litmus-portal/frontend/src/views/CreateWorkflow/WorkflowSettings/index.tsx
arkajyotiMukherjee: name.length will return either a 0 (false value) or a non zero number (true value)...so `name.length` can be used

ishangupta-ds (approved):

litmus-portal/frontend/src/views/CreateWorkflow/WorkflowSettings/index.tsx
S-ayanide: I tried using `name.length` it didn't work, when the value is 0 the if condition is if(0){} which is completely skipping the if statement

S-ayanide: if2

arkajyotiMukherjee: 0 is a false value and && is used...so a false value will short circuit the condition and if won't run...which should be the same as when length is < 0...either way it's fine by me.

arkajyotiMukherjee (approved):

PR #2780, opened 2021-05-12
(Portal - Analytics/Monitoring): Adding APIs for fetching series list and (label, values), in-memory caching, concurrent requests &  moving data pre-processing pipeline. 

ishangupta-ds:
## Proposed changes

* Adding APIs for getting series list and (label, values) from prometheus 
* Adding in-memory caching for analytics/monitoring
* Moving data pre-processing pipeline to server for metrics and annotations
* Updating sequential requests to Concurrent requests to the connected data source

## Types of changes

- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Checklist

## Dependency

- This PR will have a breaking change for analytics a frontend PR needs to be merged just after this update.

Screenshot 2021-05-12 at 1 54 59 PM

Screenshot 2021-05-12 at 8 42 54 AM

Screenshot 2021-05-12 at 8 47 54 AM

Screenshot 2021-05-12 at 8 48 12 AM

## Dependency
- PR #2787


arkajyotiMukherjee (approved):
LGTM

litmus-portal/graphql-server/graph/analytics.graphqls
```diff
@@ -108,15 +121,45 @@ input promQueryInput {
...
+type metricsPromResponse {
     queryid: String!
-    legends: [[String]]
-    tsvs: [[timeStampValue]]
+    legends: [String]
+    tsvs: [[metricsTimeStampValue]]
+}
+
+type metricsTimeStampValue{
+    date: Float
```
whitetiger1399, line 131: May we change the `date: Float` to something else

ishangupta-ds: This is to solve the 2038 problem with 32bit integer which graphql supports and mitigate frontend parsing, so we have to use float which is acceptable by visx graphs as tested. @whitetiger1399 

Ref:
https://en.wikipedia.org/wiki/Year_2038_problem
https://graphql.org/learn/schema

Screenshot 2021-05-12 at 11 00 45 AM

whitetiger1399: let it be `float` due to 32bit issue in 2k38

whitetiger1399 (approved):

litmus-portal/graphql-server/graph/analytics.graphqls
```diff
@@ -108,15 +121,49 @@ input promQueryInput {
...
+
+type annotationsPromResponse {
+    queryid: String!
+    legends: [String]
+    tsvs: [[annotationsTimeStampValue]]
+}
+
+type annotationsTimeStampValue{
+    date: Float
+    value:  Int
```
rajdas98, line 143: I think we shouldn't use Int for timestamp.

rajdas98 (changes requested):
There is too many duplicates gql schema. Can you minimize the gql schema?

rajdas98:
> There is too many duplicates gql schema. Can you minimize the gql schema?

Example-
type metricsTimeStampValue{
    date: Float    
    value:  Float
}

type annotationsTimeStampValue{
    date: Float
    value:  Int
}

ishangupta-ds:
> > There is too many duplicates gql schema. Can you minimize the gql schema?
> 
> Example-
> type metricsTimeStampValue{
> date: Float
> value: Float
> }
> 
> type annotationsTimeStampValue{
> date: Float
> value: Int
> }

we can update for the input types but changing this value type is not possible as both are required as different types on frontend (one integer and the other float) @rajdas98

litmus-portal/graphql-server/pkg/cache/cache.go
```diff
@@ -0,0 +1,22 @@
+package cache
```
rajdas98, line 1: Put the cache.go in pkg/utils directory. We are planing to move some other files to utils directory also.

litmus-portal/graphql-server/graph/analytics.graphqls
ishangupta-ds: yes we have updated to float but we cannot use string due to extra parsing overhead, flat of 64 bits may be long enough for holding the timestamp values.

ishangupta-ds:
@rajdas98 PTAL again, suggested changes are done.

rajdas98 (approved):

PR #2787, opened 2021-05-12
(Portal - Analytics/Monitoring): Removing data mapping and filtering, adding new typed queries & fixing minor style issues

ishangupta-ds:
## Proposed changes

* Removing mongoDB and prometheus data mapping from frontend
* Updating data filtration to mapping
* Adding new queries and their types for series list and (label, values)
* Fixing minor style issues for checkboxes and scrollbar

## Types of changes

## Checklist

## Dependency
- PR #2780


litmus-portal/frontend/src/components/CheckBox/index.tsx
```diff
@@ -1,9 +1,29 @@
...
+    color: theme.palette.border.main,
+    '&$checked': {
+      color: theme.palette.background.paper,
+      '& .MuiSvgIcon-root': {
+        background: theme.palette.secondary.main,
+        borderRadius: '0.25rem',
+      },
+    },
   },
+  checked: {},
```
arkajyotiMukherjee, line 14: why is this class empty?

litmus-portal/frontend/src/components/CheckBox/index.tsx
```diff
@@ -1,9 +1,29 @@
...
+    color: theme.palette.border.main,
+    '&$checked': {
+      color: theme.palette.cards.header,
+      '& .MuiSvgIcon-root': {
+        background: theme.palette.secondary.main,
+        borderRadius: '0.25rem',
+      },
+    },
+  },
+  checked: {},
```
arkajyotiMukherjee, line 28: same

litmus-portal/frontend/src/pages/MonitoringDashboardPage/styles.ts
```diff
@@ -218,6 +218,8 @@ export const useOutlinedInputStyles = makeStyles((theme: Theme) => ({
     color: theme.palette.text.hint,
     background: theme.palette.background.paper,
   },
+  focused: {},
+  notchedOutline: {},
```
arkajyotiMukherjee, line 222: same

litmus-portal/frontend/src/components/CheckBox/index.tsx
ishangupta-ds: Without adding this, while overriding the base styles the `&$checked` selector does not work anymore for this MUI component. @arkajyotiMukherjee

litmus-portal/frontend/src/components/CheckBox/index.tsx
ishangupta-ds: same as above

litmus-portal/frontend/src/pages/MonitoringDashboardPage/styles.ts
ishangupta-ds: same as above

arkajyotiMukherjee (approved):

litmus-portal/frontend/src/components/CheckBox/index.tsx
whitetiger1399: Do we require to have two components `CheckBoxBlue` and `CheckBoxWhite`? May we please have only one component and override the styles for the other. Or we don't create the local component and simply pass the styles as required.

litmus-portal/frontend/src/views/AnalyticsDashboard/DataSource/Forms/prometheus.tsx
```diff
@@ -306,7 +301,7 @@ const ConfigurePrometheus: React.FC<ConfigurePrometheusProps> = ({
             <FormControlLabel
               className={classes.withCACert}
               control={
-                <Checkbox
+                <CheckBoxWhite
```
whitetiger1399, line 304: same

litmus-portal/frontend/src/components/CheckBox/index.tsx
ishangupta-ds: Yes it should be from MUI for now we can use local one, also the two components were required here because of different colour properties which cannot be passed on class and have to be overridden on MUI component. (when using MUI we can shift to using a single one with proper style interface, workflow analytics team is already working on the same) this PR is just for fixing the colour.

arkajyotiMukherjee: try passing the color as a prop to your component....withStyles is a HOC it should have an API to pass props @ishangupta-ds

whitetiger1399: Using the checkbox compoent in (https://github.com/whitetiger1399/litmus-ui/tree/autoCompleteChip).  
Screenshot from 2021-05-12 18-10-10
Screenshot from 2021-05-12 18-10-22 
We may pass a className which will override the default m-ui properties as per your requirement

ishangupta-ds: Tried out the above approach @whitetiger1399 , it does not work for the selectors for base class properties its working fine like in above screenshot as they are styles for root `checked` and `MuiSvgIcon-root` are not getting overridden, so only colour worked from the class mentioned below but we need both of them as an override. Please try to check if these are working for you in litmus-ui for select chip.

  rootCheck: {
    color: theme.palette.border.main,
    '&$checked': {
      color: theme.palette.background.paper,
      '& .MuiSvgIcon-root': {
        background: theme.palette.secondary.main,
        borderRadius: '0.25rem',
      },
    },
  },
  checkedCheck: {},
  
  
Screenshot 2021-05-12 at 7 40 18 PM

ishangupta-ds: @arkajyotiMukherjee makeStyle allows props directly, we can use useStyles but then it no longer stays a pure react component so it cannot be used in jsx directly unless we extend the Checkbox from MUI for this its better to do it in litmus-ui than in litmus.

useStyles:

export const StyleWithThemeProps = (props: any) => {
  return withStyles(
    (theme) => ({
      root: {
        color: theme.palette.border.main,
        '&$checked': {
          color: props.checkedColor,
          '& .MuiSvgIcon-root': {
            background: theme.palette.secondary.main,
            borderRadius: '0.25rem',
          },
        },
      },
      checked: {},
    }),
    { withTheme: true }
  )(Checkbox);
};

makeStyles example:

const useStyles = makeStyles({
  root: {
    background: props => props.color,
    "&:hover": {
      background: props => props.hover
    },
    border: 0,
    borderRadius: 3,
    color: "white",
    height: 48,
    padding: "0 30px"
  }
});

export default function Hook() {
  const classes = useStyles({
    color: "red",
    hover: "blue"
  });
  return <Button className={classes.root}>Hook</Button>;
}

ishangupta-ds: For now till we use from litmus-ui we may need to use these two styled/overrided components separately because MUI is not offering a direct way to override the css selectors without overriding the base styles by wrapping it up with the withStyles HOC @arkajyotiMukherjee @whitetiger1399

litmus-portal/frontend/src/containers/layouts/Scaffold/styles.ts
```diff
@@ -12,14 +12,14 @@ const useStyles = makeStyles((theme: Theme) => ({
     gridTemplateAreas: '"header header" "sidebar content"',
 
     '& ::-webkit-scrollbar': {
-      width: '0.4rem',
+      width: '0.35rem',
     },
     '& ::-webkit-scrollbar-track': {
-      marginTop: theme.spacing(1),
       webkitBoxShadow: `inset 0 0 8px ${theme.palette.common.black}`,
```
vanshBhatia-A4k9, line 18: This was added to avoid the scrollbar merging into the header as the color of the header has changed now, scrolling all the way to the top makes it so that the scrollbar merges into the header

litmus-portal/frontend/src/containers/layouts/Scaffold/styles.ts
```diff
@@ -12,14 +12,14 @@ const useStyles = makeStyles((theme: Theme) => ({
     gridTemplateAreas: '"header header" "sidebar content"',
 
     '& ::-webkit-scrollbar': {
-      width: '0.4rem',
```
vanshBhatia-A4k9, line 15: We are already increasing the font size in buttons 14px->16px and the font style to Inter to make things more visible..would it be a good idea to reduce the size of the scrollbar...?

litmus-portal/frontend/src/containers/layouts/Scaffold/styles.ts
ishangupta-ds: Yes i checked this but with the white color on top it looks like we have more area to scroll up to, may be a scroll bar track color can solve the problem if this is the case. @vanshBhatia-A4k9

ishangupta-ds: Its also overriding all the litmus ui graph legend table scrollbars, modal/pop over scrollbars and yaml editor scrollbar

litmus-portal/frontend/src/containers/layouts/Scaffold/styles.ts
ishangupta-ds: The problem is doing this overriding all scrollbars graph legend tables etc, we need to override either in litmus-ui and all child components getting affected or inject this scrollbar css component/page wise where required @vanshBhatia-A4k9

litmus-portal/frontend/src/components/CheckBox/index.tsx
ishangupta-ds: Ignore the  comments @arkajyotiMukherjee and @whitetiger1399 the props method worked !

ishangupta-ds:
@mallick370 @arkajyotiMukherjee @vanshBhatia-A4k9 @whitetiger1399 can we decide on the styles for the scrollbar we have 3 currently, also where to override the same (in litmus-ui or individual components/pages)

Screenshot 2021-05-13 at 8 36 17 AM

Screenshot 2021-05-13 at 8 36 10 AM

Screenshot 2021-05-13 at 8 45 13 AM

vanshBhatia-A4k9:
> @mallick370 @arkajyotiMukherjee @vanshBhatia-A4k9 @whitetiger1399 can we decide on the styles for the scrollbar we have 3 currently, also where to override the same (in litmus-ui or individual components/pages)
> 
> Screenshot 2021-05-13 at 8 36 17 AM
> 
> Screenshot 2021-05-13 at 8 36 10 AM
> 
> Screenshot 2021-05-13 at 8 45 13 AM

I believe a smaller margin (2px below) might be better?
Screenshot from 2021-05-13 08-49-34

litmus-portal/frontend/src/components/CheckBox/index.tsx
ishangupta-ds: Although it worked we still have the same issue that the overriding failed after using props.

Screenshot 2021-05-13 at 8 57 01 AM

whitetiger1399 (approved):

ishangupta-ds:
> > @mallick370 @arkajyotiMukherjee @vanshBhatia-A4k9 @whitetiger1399 can we decide on the styles for the scrollbar we have 3 currently, also where to override the same (in litmus-ui or individual components/pages)
> > Screenshot 2021-05-13 at 8 36 17 AM
> > Screenshot 2021-05-13 at 8 36 10 AM
> > Screenshot 2021-05-13 at 8 45 13 AM
> 
> I believe a smaller margin (2px below) might be better?
> Screenshot from 2021-05-13 08-49-34

yes we will be overriding this at component level as discussed. @vanshBhatia-A4k9

Jonsy13 (approved):

vanshBhatia-A4k9 (approved):

PR #2829, opened 2021-05-21
Enhancement: Moving pagination, sorting and filtering of workflow runs table to the backend

arkajyotiMukherjee:
## Proposed changes
Improvements:
- adding pagination to workflowRuns in the database queries
- moving all the filtering and sorting to the db queries
- adding some commonly used metadata from executionData to workflowRun schema so that we can have reduced network load in the frontend like phase, resiliency score etc.
- added frontend changes to use the new API.

Contains both frontend and backend changes

## Types of changes

## Checklist

## Dependency


litmus-portal/graphql-server/graph/schema.graphqls
```diff
@@ -6,7 +6,7 @@ directive @authorized on FIELD_DEFINITION
 
 type Cluster {
   cluster_id: ID!
-  project_id: ID!
+  project_id: String!
```
gdsoumya, line 9: Why is this string? Should be consistent with other id;s

litmus-portal/graphql-server/pkg/chaos-workflow/handler/handler.go
```diff
@@ -115,37 +116,191 @@ func UpdateWorkflow(ctx context.Context, input *model.ChaosWorkFlowInput, r *sto
...
+		if endIndex > len(result) {
+			endIndex = len(result)
+		}
+		result = result[startIndex:endIndex]
+	}
+
+	output := model.GetWorkflowsOutput{
+		TotalNoOfWorkflowRuns: totalNoOfRuns,
+		WorkflowRuns:          result,
+	}
```
gdsoumya, line 299: Filtering not optimised will be a very slow query. All the filters should be at db level.if we are manually doing thee filters there is no benefit of moving it to the backend

gdsoumya (changes requested):

litmus-portal/graphql-server/pkg/chaos-workflow/handler/handler.go
SarthakJain26: Due to the current structure of DB, we were bind to go with this approach. To shift all operations at DB level, we will need another schema consisting of only workflow runs having workflow_id as a foreign key. But due to some reasons it was decided to not split the schemas and go with this approach.
The benefit of moving it to the backend is that it will reduce the network load. Previously all these operations were done in frontend which was even more unoptimized from an end user perspective.

gdsoumya: Shifting to backend doesn't mean it will make things more efficient, in a setup where there is only 1 gql server (current scenario) fo reach request serrver has to do so much computation(searching o(n)+ sorting o(nlogn) or possibly o(n^2)) and if there are 100 users querying for 100 workflows each the amount of time taken to process each request increases this locks up server resources. In a small scale setup this won't be visible but when there is more load/workflows it will take a toll. Either we make this more optimized or wait until the server can be scaled.

SarthakJain26: I already told you about the current scenario. Do you have any better approach to this problem than the one that we have used? How can we make it more optimized?

gdsoumya: The current scenario will not be solved by this solution, I can assure you that, slower queries won't be any better that more network load. 
Will have to look into this later, currently have other urgent tasks. Better to postpone this, there is already a task going on to scale the server if by then you can't figure any optimised solution we can use it. Also if dividing into 2 separate collections makes this easier you can discuss that approach too, I personally don't see any problem there.

rajdas98: I think a few days back, we have calculated the DB calls after dividing them into two collections. It will take more DB calls than the current implementation.

rajdas98: The other benefit of this approach is the computation done in the frontend will be removed. I agree on that part that it will definitely slow the queries/response time.

rajdas98: Try to see a join or Aggregated mongo query for this.

litmus-portal/graphql-server/graph/schema.graphqls
arkajyotiMukherjee: In most places it is String....in some it was ID...so I changed it to String.
If you want we can make everything to IDs as well but we need something consistent.
@gdsoumya

gdsoumya: Sure you can change to either but the other id types like `cluster_id` etc. should also be converted to the same type then

litmus-portal/graphql-server/pkg/chaos-workflow/handler/handler.go
arkajyotiMukherjee: The best approach according to me which we gave initially was to split it into different collections. I too want to shift it to the db layer but the current schema is not suitable for that. I'll have a solution in mind...will do some research and try that.
@gdsoumya @rajdas98

gdsoumya: Currently, I don't see any lag in the frontend for the browse workflow section even when we have a considerable amount of workflows. With further increase in workflow count it might start lagging but the effect on the sever will be more prominent with this solution. if the server starts to slow down for each query being made more resources are being held up that would mean less number of requests can be served. It not just slows the current request but also can lead to a compounding effect on all other requests. 
Also if we shift the searching/sorting to the server that would mean the frontend will make a network call/request to the server anytime a user wants to filter the workflows. Previously it was done on the frontend so it could possibly reuse the existing data from the previous call itself. This would also mean now we have more network load, specially because we are providing realtime search results - every time a user types a letter we filter the results, if this uses the backend that would mean each key press is a network call

gdsoumya: This problem needs further discussion and proper design if we want to do this right.

litmus-portal/graphql-server/graph/schema.graphqls
arkajyotiMukherjee: Yes, that's the problem...it's all over the place right now....I'm going with ID then in all places @gdsoumya @rajdas98

litmus-portal/frontend/src/views/ChaosWorkflows/BrowseWorkflow/index.tsx
```diff
@@ -16,74 +16,134 @@ import ExpandMoreIcon from '@material-ui/icons/ExpandMore';
...
+    cluster_name: 'All',
+    workflow_status: 'All',
+    date_range: {
+      start_date: new Date(0).valueOf().toString(),
+    },
+  });
+
+  // State for date to be displayed
+  const [displayDate, setDisplayDate] = React.useState<string>(
+    'Select a period'
```
ishangupta-ds, line 64: Translation can be added.

litmus-portal/graphql-server/graph/schema.graphqls
```diff
@@ -428,16 +423,25 @@ type Mutation {
...
+  ): ImageRegistryResponse! @authorized
+
+  updateImageRegistry(
+    image_registry_id: String!
+    project_id: String!
+    imageRegistryInfo: imageRegistryInput!
+  ): ImageRegistryResponse! @authorized
+
+  deleteImageRegistry(image_registry_id: String!, project_id: String!): String!
+    @authorized
```
ishangupta-ds, line 444: Do a `gofmt`

ishangupta-ds (changes requested):
One translation can be added, `gofmt` too. Also generated binary can be removed from the server. @arkajyotiMukherjee

litmus-portal/graphql-server/pkg/database/mongodb/workflow/schema.go
```diff
@@ -21,20 +25,98 @@ type ChaosWorkFlowInput struct {
...
+	Phase             string   `bson:"phase"`
+	ResiliencyScore   *float64 `bson:"resiliency_score,string,omitempty"`
+	ExperimentsPassed *int     `bson:"experiments_passed,string,omitempty"`
+	TotalExperiments  *int     `bson:"total_experiments,string,omitempty"`
+	ExecutionData     string   `bson:"execution_data"`
+	Completed         bool     `bson:"completed"`
+}
+
+type ExecutionData struct {
+	WorkflowID        string          `json:"-"`
```
rajdas98, line 53:
```suggestion
	WorkflowID        string          `bson:"-"`
```

please change this to other places

rajdas98 (changes requested):
Please remove this bin- litmus-portal/graphql-server/server

litmus-portal/graphql-server/pkg/database/mongodb/workflow/schema.go
arkajyotiMukherjee: these are used for parsing the ExecutionData...we are storing it the way it was before as a string...hence json. we moved it here as in future PRs this struct will be shared. It was previously inside operations.go. @rajdas98

ishangupta-ds: If its specific for just the operations, you can move it to workflow types @arkajyotiMukherjee

gdsoumya: Yes if it isn't related to database it shouldn't be in the db pkg

gdsoumya: Also this is not updated, there have been changes to subscriber and this needs to be updated too

gdsoumya (changes requested):

arkajyotiMukherjee:
Addressed all changes. PTAL @rajdas98 @gdsoumya @ishangupta-ds @Jonsy13

ishangupta-ds (approved):
LGTM

ishangupta-ds:
@arkajyotiMukherjee please check the import order some checks are failing

arkajyotiMukherjee:
done @ishangupta-ds

ishangupta-ds (approved):

gdsoumya (approved):
Changes pending:
- optimise the pipeline and filter in one step

Jonsy13 (approved):

rajdas98 (approved):

Issue #2483, opened 2021-03-01
Authentication Module Refactor


ajeshbaby:
- Rewrite/Refactor the authentication server to make it lightweight and modular so that 
  - It can be moved into a separate repository
  - Handle different authentication modules
      -  Local
      -  Gmail Auth (Mandatory)
      -  Github (Optional) 
 - Need a feature to reset the existing password.

Debanitrkl:
@rajdas98 @ajeshbaby I'm interested in this project. Could you help me with some resources to begin

specter25:
@rajdas98 hi  sir i am interested to contribute to this project in GSoC 2021 . Could you provide me with some resources . I did study the code of the authentication server ( https://github.com/litmuschaos/litmus/tree/master/litmus-portal/authentication ) . I believe the gmail auth and google auth is yet not implemented .

Debanitrkl:
> Hi @Debanitrkl, This issue is locked for GSoC 2021, Feel free to send a proposal once GSoC announces this project.

Yes sir I'm interested for contributing through GSoC, would start writing my proposal soon, just wanted to get started with the code base of authentication server

specter25:
@rajdas98 the current auth setup uses "mgo" go package which is not well maintained as compared to the official go.mongodb.org/mongo-driver/mongo driver , should we prefer using the official mongo-driver in the new api ?

gdsoumya:
Yes you have to use the official mongo go driver

specter25:
@rajdas98 @gdsoumya , sir one more doubt , the current implementation uses mongodb for session management but i believe using redis is a better solution for session management . So should we go with mongodb or switch to redis for managing sessions ?

gdsoumya:
There is no session management I believe, we use jwt tokens for authorization so session management in the true sense is not needed in our case.

specter25:
okay sir thanks .

rajdas98:
@specter25 , Currently, we are not planning to use other db. The goal of this project is to make the auth-server light-weight and add other thirty party integration.

specter25:
Okay sir got it .

DarthBenro008:
@rajdas98 by light-weight, in what aspects is it expected to be light? lesser third party dependencies? or more memory efficient paradigms of code? A quick gist would be insightful!

rajdas98:
[...]

PR #2867, opened 2021-06-07
refactor: litmus-portal authentication server

DarthBenro008:
# Authentication Server Refactor

> Part of #2483 and [GSoC Project](https://summerofcode.withgoogle.com/projects/#5058541729087488)

This PR comprehends a full re-write of litmus-portal's golang authentication server.

The server has been developed keeping in mind that the litmus-portal frontend does not need to make any change in request/response.

## Routes Included

| Type of Request 	|      Routes      	|               Functionality              	| Authorisation Required 	|
|-----------------	|:----------------:	|:----------------------------------------:	|------------------------	|
| POST            	| /login           	| Logs in the user and generates JWT Token 	| No                     	|
| GET             	| /users           	| Fetches all the users to be displayed    	| Yes                    	|
| POST            	| /create          	| Allows admin to create a user            	| Yes (with admin role)  	|
| POST            	| /update/password 	| Update password of logged in user        	| Yes                    	|
| POST            	| /reset/password  	| Allows admins to reset user's password   	| Yes (with admin role)  	|
| POST            	| /update/details  	| Allows user to update email and name     	| Yes                    	|

## Dependencies Used

```go
go 1.16
require (	
        github.com/dgrijalva/jwt-go v3.2.0+incompatible	
        github.com/gin-contrib/cors v1.3.1	
        github.com/gin-gonic/gin v1.7.2	
        github.com/sirupsen/logrus v1.4.2	
        go.mongodb.org/mongo-driver v1.5.3	
        golang.org/x/crypto v0.0.0-20210513164829-c07d793c2f9a
)
```

- [gin-gionic](https://github.com/gin-gonic/gin) 
- [mongo-driver](https://github.com/mongodb/mongo-go-driver)
- [logrus](https://github.com/sirupsen/logrus)
- [jwt-go](https://github.com/dgrijalva/jwt-go)

## Results

> The current authentication server and the re-written server were deployed on `Okteto` and tested

### Performance Gain

The current authentication server took **14 seconds** for an average response

litmus_current_auth_marked

The re-written authentication server took **0.3 seconds** for an average response

marked_new

> This is a boost in performance of almost `46 times` meanwhile keeping the functionality and the security the same.

### Easier Packaging of Code

The packages and folders have been completely simplified. Everything in the codebase has been decoupled for easier development and access.

Before: 
old_directory

After:
new_directory

## Try out

You can find this codebase deployed on https://litmusportal-production-frontend-service-darthbenro008.cloud.okteto.net/ 
Username: `admin`
Password: `litmus`

Alternatively, you can also deploy this codebase on your own `Okteto Cluster` by [clicking here](https://cloud.okteto.com/#/deploy?repository=https://github.com/DarthBenro008/litmus&branch=okteto-testing) 

## ToDo
- Unit Testing and DB Mocking


litmus-portal/authentication/api/routes/user_router.go
```diff
@@ -0,0 +1,184 @@
...
+		err := c.BindJSON(&userRequest)
+		if err != nil {
+			log.Warn(err)
+			c.JSON(utils.ErrorStatusCodes[utils.ErrInvalidRequest], presenter.CreateErrorResponse(utils.ErrInvalidRequest))
+			return
+		}
+		if userRequest.UserName == "" || userRequest.Password == "" {
+			c.JSON(utils.ErrorStatusCodes[utils.ErrInvalidRequest], presenter.CreateErrorResponse(utils.ErrInvalidRequest))
+			return
+		}
```
SarthakJain26, line 98: Better to sanitize the data before authenticating, so that trailing spaces could be trimmed

litmus-portal/authentication/api/routes/user_router.go
```diff
@@ -0,0 +1,184 @@
...
+		err := c.BindJSON(&userRequest)
+		if err != nil {
+			log.Warn(err)
+			c.JSON(utils.ErrorStatusCodes[utils.ErrInvalidRequest], presenter.CreateErrorResponse(utils.ErrInvalidRequest))
+			return
+		}
+		if userRequest.Role == "" || userRequest.UserName == "" || userRequest.Password == "" || userRequest.Email == "" {
+			c.JSON(utils.ErrorStatusCodes[utils.ErrInvalidRequest], presenter.CreateErrorResponse(utils.ErrInvalidRequest))
+			return
+		}
```
SarthakJain26, line 43: Better to sanitize the data before creating, so that trailing spaces could be trimmed

litmus-portal/authentication/api/routes/user_router.go
```diff
@@ -0,0 +1,184 @@
...
+		err = service.UpdatePassword(&userPasswordRequest, false)
+		if err != nil {
+			log.Error(err)
+			c.AbortWithStatusJSON(utils.ErrorStatusCodes[utils.ErrServerError], presenter.CreateErrorResponse(utils.ErrServerError))
+			return
+		}
+		c.JSON(200, gin.H{
+			"message": "password has been reset successfully",
```
SarthakJain26, line 181: Typo: suddenly -> successfully

litmus-portal/authentication/pkg/entities/user.go
```diff
@@ -0,0 +1,67 @@
...
+
+	"github.com/dgrijalva/jwt-go"
+	"github.com/sirupsen/logrus"
+	"go.mongodb.org/mongo-driver/bson/primitive"
+)
+
+//User contains the user information
+type User struct {
+	ID        primitive.ObjectID `bson:"_id,omitempty" json:"_id,omitempty"`
+	UserName  string             `bson:"username,omitempty" json:"username,omitempty"`
```
SarthakJain26, line 15: Should we use omitempty with ID, UserName, Role and Password? These fields will/should never be nil :thinking:

litmus-portal/authentication/pkg/entities/user.go
```diff
@@ -0,0 +1,67 @@
...
+	Role      Role               `bson:"role,omitempty" json:"role,omitempty"`
+	LoggedIn  bool               `bson:"logged_in,omitempty" json:"logged_in,omitempty"`
+	CreatedAt *time.Time         `bson:"created_at,omitempty" json:"created_at,omitempty"`
+	UpdatedAt *time.Time         `bson:"updated_at,omitempty" json:"updated_at,omitempty"`
+	RemovedAt *time.Time         `bson:"removed_at,omitempty" json:"removed_at,omitempty"`
+}
+
+//UserPassword defines structure for password related requests
+type UserPassword struct {
+	Username    string `json:"username,omitempty"`
```
SarthakJain26, line 28: Same as above

litmus-portal/authentication/api/routes/user_router.go
```diff
@@ -0,0 +1,184 @@
...
+)
+
+//UserRouter creates all the required routes for user authentications purposes.
+func UserRouter(router *gin.Engine, service user.Service) {
+	router.POST("/login", loginUser(service))
+	router.Use(middleware.JwtMiddleware())
+	router.POST("/update/password", updatePassword(service))
+	router.POST("/reset/password", resetPassword(service))
+	router.POST("/create", createUser(service))
+	router.POST("/update/details", updateUser(service))
```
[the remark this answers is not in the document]
DarthBenro008: Changing the protocol will required changes from the frontend and as mentioned above, the server was re-written keeping in mind that there should be no changes in request/response as in the current authentication-server used by litmus-frontend. I would prefer PUT over POST. Had to choose PUT to avoid further changes in litmus-frontend.

litmus-portal/authentication/api/main.go
```diff
@@ -0,0 +1,79 @@
...
+	app.Use(cors.New(config))
+	routes.UserRouter(app, userService)
+	err = app.Run(PORT)
+	if err != nil {
+		log.Fatalf("Failure to start litmus-portal authentication server due to %s", err)
+	}
+}
+
+func validatedAdminSetup(service user.Service) {
+	configs := map[string]string{"ADMIN_PASSWORD": utils.AdminPassword, "ADMIN_USERNAME": utils.AdminName, "DB_USER": utils.DBUser, "DB_SERVER": utils.DBUrl, "DB_NAME": utils.DBName, "DB_PASSWORD": utils.DBPassword, "JW_SECRET": utils.JwtSecret}
```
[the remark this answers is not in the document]
DarthBenro008: Will amend this.

litmus-portal/authentication/pkg/entities/user.go
DarthBenro008: - We do not want to send Password while sending responses for `/users` GET route, hence omitempty is required.
- I agree ID, UserName, Role are never be nil and is guaranteed from the server's side as there are validations and checks.
- but `omitempty` is significant only when we are de-serialising the json/bson. Hence I think there is no difference if we write or do not write `omitempty`

litmus-portal/authentication/pkg/entities/user.go
DarthBenro008: - The `/reset/password` POST route does not send a `Username`, hence `omitempty` is required

litmus-portal/authentication/api/routes/user_router.go
DarthBenro008: I do not seem to find the mentioned typo anywhere 😅

litmus-portal/authentication/api/routes/user_router.go
DarthBenro008: The `litmus-frontend` already does this. As this a specific use-case API (only used by litmus-frontend), we can choose to avoid sanitisation of input data in favour to not add an additional dependency.

DarthBenro008:
Hey @SarthakJain26 , Thank you for the comments. I have addressed/resolved the same.

litmus-portal/authentication/api/routes/user_router.go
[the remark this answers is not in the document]
ishangupta-ds: @DarthBenro008 its better to change this cc: @rajdas98 @gdsoumya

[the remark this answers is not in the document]
DarthBenro008: @ishangupta-ds , I would like to confirm,  both the `/update/details` and `/update/password` routes would have `PUT` requests if we will be going ahead with it? 

## New Routes Table

| Type of Request 	|      Routes      	|               Functionality              	| Authorisation Required 	|
|-----------------	|:----------------:	|:----------------------------------------:	|------------------------	|
| POST            	| /login           	| Logs in the user and generates JWT Token 	| No                     	|
| GET             	| /users           	| Fetches all the users to be displayed    	| Yes                    	|
| POST            	| /create          	| Allows admin to create a user            	| Yes (with admin role)  	|
| PUT            	| /update/password 	| Update password of logged in user        	| Yes                    	|
| POST            	| /reset/password  	| Allows admins to reset user's password   	| Yes (with admin role)  	|
| PUT            	| /update/details  	| Allows user to update email and name     	| Yes                    	|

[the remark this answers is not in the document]
ishangupta-ds: Ideally both should be updated @DarthBenro008 , awaiting comments from @rajdas98 and @gdsoumya too.

litmus-portal/authentication/api/routes/user_router.go
```diff
@@ -0,0 +1,184 @@
...
+	router.GET("/users", fetchUsers(service))
+}
+
+func createUser(service user.Service) gin.HandlerFunc {
```
ishangupta-ds, line 26: Handlers / controllers like this could have been placed in a separate file, this can be kept just for all the routes. @DarthBenro008

litmus-portal/authentication/pkg/entities/user.go
```diff
@@ -0,0 +1,67 @@
...
+
+//GetSignedJWT generates the JWT Token for the user object
+func (user *User) GetSignedJWT() (string, error) {
+
+	token := jwt.New(jwt.SigningMethodHS512)
+	claims := token.Claims.(jwt.MapClaims)
+	claims["uid"] = user.ID.Hex()
+	claims["role"] = user.Role
+	claims["username"] = user.UserName
+	claims["exp"] = time.Now().Add(time.Minute * 300).Unix()
```
ishangupta-ds, line 58: should it not be for 24 hours?

litmus-portal/authentication/pkg/user/repository.go
```diff
@@ -0,0 +1,140 @@
...
+type repository struct {
+	Collection *mongo.Collection
+}
+
+//FindUser helps to authenticate the user
+func (r repository) FindUser(user *entities.User) (*entities.User, error) {
+	var result = entities.User{}
+	findOneErr := r.Collection.FindOne(context.TODO(), bson.M{
+		"username": user.UserName,
+	}).Decode(&result)
```
ishangupta-ds, line 33: Have you formatted using `gofmt` ?

litmus-portal/authentication/api/main.go
```diff
@@ -0,0 +1,79 @@
...
+
+	flag.Parse()
+	// Version Info
+	printVersion()
+
+	db, err := utils.DatabaseConnection()
+	if err != nil {
+		log.Fatal("Database Connection Error $s", err)
+	}
+	err = utils.CreateIndex("users", "username", db)
```
ishangupta-ds, line 35: Declare the collection name and field name as constants, then use.

litmus-portal/authentication/api/main.go
```diff
@@ -0,0 +1,79 @@
...
+	userCollection := db.Collection("users")
```
ishangupta-ds, line 36: same for this

ishangupta-ds (changes requested):

ishangupta-ds:
@DarthBenro008 there is a merge conflict too.

litmus-portal/authentication/pkg/user/repository.go
DarthBenro008: Yes, the whole project has been linted using `gofmt`

litmus-portal/authentication/pkg/entities/user.go
DarthBenro008: Oh, yes. It ideally should be 24 hours. Should we allow the user to set this via `env` or should we just change the multiplies to 1440 from 300?

ishangupta-ds: Use an env @DarthBenro008

litmus-portal/authentication/api/main.go
```diff
@@ -0,0 +1,79 @@
...
+	"litmus/litmus-portal/authentication/pkg/user"
+	"litmus/litmus-portal/authentication/pkg/utils"
+	"runtime"
+
+	"github.com/gin-contrib/cors"
+	"github.com/gin-gonic/gin"
+	log "github.com/sirupsen/logrus"
+)
+
+var (
```
ishangupta-ds, line 17: Const

ishangupta-ds (changes requested):

litmus-portal/authentication/api/routes/user_router.go
DarthBenro008: I do not want to export these functions (handlers) and make sure it's available under the same package and to the API only. The code architecture is highly inspired by Clean Architecture. Maybe the package name `routes` is misleading and could be changed to `handlers` instead?

ishangupta-ds: @DarthBenro008 by clean arch. we define the handler as a service, but in this case you have created a service file which has all component based services. There are two options - 

* Either you can create a custom middleware for the services if you feel these functions cannot be a part of the service routines.

*  If they can be then i would suggest to move this logic to the services itself and make them proper handlers.

Any approach can be taken but routing package should not have the handler logic or the middleware logic. It can be kept as a simple router with mapping of routes and their corresponding handlers.

ishangupta-ds: Changing the package name to handlers will be confusing for the people looking for the routing map in future.

litmus-portal/authentication/api/routes/user_router.go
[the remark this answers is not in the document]
rajdas98: IMO, If we use query params in `/update/details` and  `/update/password`, then having `PUT` method makes sense. Otherwise, It will not make any difference if we use  `POST`.

litmus-portal/authentication/api/routes/user_router.go
rajdas98: I think there will be a problem if we remove trailing spaces from the password in the backend. Example- `pass  word` and `password` both will pass, if we remove trailing spaces.

litmus-portal/authentication/pkg/entities/user.go
rajdas98: IMO. It's better to remove `omitempty` if the fields are compulsory.

litmus-portal/authentication/api/routes/user_router.go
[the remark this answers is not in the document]
SarthakJain26: Let's leave it as a POST request for now, we can update it later along with the frontend.

litmus-portal/authentication/api/routes/user_router.go
SarthakJain26: Screenshot from 2021-06-09 09-57-57

IDK how its happening, but my system autocorrects successful-ly to suddenly :sweat_smile:

litmus-portal/authentication/api/routes/user_router.go
SarthakJain26: IMO we should atleast sanitise/validate username. Currently frontend is not handling this

rajdas98: Yes, that can be done

DarthBenro008: image

image



I assumed by validation, we are talking solely about trailing spaces, which are handled by the front-end as shown above. Is there any additional regex/username constraints we need to abide by?

litmus-portal/authentication/api/routes/user_router.go
DarthBenro008: @ishangupta-ds given the options, I think I will take the Controller approach (i.e an API oriented package rather than a service-oriented package as those functions as Gin Handlers). I will call these "controllers" as "handlers" in the codebase as they as Gin Handler funcs after all. I do not want to make these service middlewares as they depend on the API that handles the request.

An example of UserRouter definition after the change:

```go
func UserRouter(router *gin.Engine, service user.Service) {
	router.POST("/login", handlers.LoginUser(service))
	router.Use(middleware.JwtMiddleware())
	router.POST("/update/password", handlers.UpdatePassword(service))
	router.POST("/reset/password", handlers.ResetPassword(service))
	router.POST("/create", handlers.CreateUser(service))
	router.POST("/update/details", handlers.UpdateUser(service))
	router.GET("/users", handlers.FetchUsers(service))
}
```

litmus-portal/authentication/api/routes/user_router.go
SarthakJain26: There should not be spaces at the end. "admin " should be trimmed to "admin"

DarthBenro008: Ah, yes. We will need to add sanitisation. I was thinking should I create a middleware giving validation as well as sanitisation definitions for each route and pass it in the router. Something similar to [Joi](https://www.npmjs.com/package/joi) in the NodeJS world? Or would it be overkill for this project?

cc: @rajdas98 @gdsoumya @ishangupta-ds

litmus-portal/authentication/api/routes/user_router.go
ishangupta-ds: Sure !

DarthBenro008:
@ishangupta-ds @SarthakJain26 @rajdas98 All the requested changes have been addressed and resolved :smiley:

litmus-portal/authentication/pkg/utils/configs.go
```diff
@@ -2,14 +2,24 @@ package utils
...
-	DBPassword    = os.Getenv("DB_PASSWORD")
-	DBName        = "auth"
+	JwtSecret         = os.Getenv("JWT_SECRET")
+	AdminName         = os.Getenv("ADMIN_USERNAME")
+	AdminPassword     = os.Getenv("ADMIN_PASSWORD")
+	DBUrl             = os.Getenv("DB_SERVER")
+	DBUser            = os.Getenv("DB_USER")
+	DBPassword        = os.Getenv("DB_PASSWORD")
+	JWTExpiryDuration = getEnvAsInt("JWT_EXPIRY_MINS", 1440)
+	DBName            = "auth"
```
ishangupta-ds, line 16: Why is this DBName in config and the collection and field names in the main file? Also the port can be moved to config @DarthBenro008

ishangupta-ds (changes requested):

litmus-portal/authentication/pkg/utils/configs.go
DarthBenro008: Oops :sweat_smile: , I had taken the reference from the current authentication server, hence followed the same. It has been moved to configs @ishangupta-ds

litmus-portal/authentication/pkg/utils/configs.go
```diff
@@ -0,0 +1,29 @@
...
+	Port              = ":3000"
+	CollectionName    = "users"
+
+	UsernameField = "username"
```
ishangupta-ds, line 20: Do a ‘gofmt’ and also remove the extra line.

ishangupta-ds (changes requested):

litmus-portal/authentication/pkg/utils/configs.go
DarthBenro008: Done!

ishangupta-ds (approved):
LGTM

SarthakJain26 (approved):

rajdas98 (approved):

litmus-portal/authentication/api/handlers/user_handlers.go
```diff
@@ -0,0 +1,176 @@
...
+		if err != nil {
+			log.Warn(err)
+			c.JSON(utils.ErrorStatusCodes[utils.ErrInvalidRequest], presenter.CreateErrorResponse(utils.ErrInvalidRequest))
+			return
+		}
+		if userRequest.Role == "" || userRequest.UserName == "" || userRequest.Password == "" {
+			c.JSON(utils.ErrorStatusCodes[utils.ErrInvalidRequest], presenter.CreateErrorResponse(utils.ErrInvalidRequest))
+			return
+		}
+		userRequest.UserName = utils.SanitizeString(userRequest.UserName)
```
gdsoumya, line 33: Trimming should happen before empty check, the empty check will allow " " data which would imply empty data after trimming.

gdsoumya (changes requested):

litmus-portal/authentication/api/handlers/user_handlers.go
```diff
@@ -0,0 +1,176 @@
...
+		if err != nil {
+			log.Warn(err)
+			c.JSON(utils.ErrorStatusCodes[utils.ErrInvalidRequest], presenter.CreateErrorResponse(utils.ErrInvalidRequest))
+			return
+		}
+		if userRequest.UserName == "" || userRequest.Password == "" {
+			c.JSON(utils.ErrorStatusCodes[utils.ErrInvalidRequest], presenter.CreateErrorResponse(utils.ErrInvalidRequest))
+			return
+		}
+		userRequest.UserName = utils.SanitizeString(userRequest.UserName)
```
gdsoumya, line 89: Same

litmus-portal/authentication/api/handlers/user_handlers.go
```diff
@@ -0,0 +1,176 @@
...
+		adminUser.ID, _ = primitive.ObjectIDFromHex(uid)
+		if err != nil {
+			log.Warn(err)
+			c.JSON(utils.ErrorStatusCodes[utils.ErrInvalidRequest], presenter.CreateErrorResponse(utils.ErrInvalidRequest))
+			return
+		}
+		if userPasswordRequest.OldPassword == "" || userPasswordRequest.NewPassword == "" {
+			c.JSON(utils.ErrorStatusCodes[utils.ErrInvalidRequest], presenter.CreateErrorResponse(utils.ErrInvalidRequest))
+			return
+		}
```
gdsoumya, line 129: Should we put the password length/chars verification to the backend? " " password is a valid one according to this. Depending on frontend is not okay because we might have CLIs and external integrations so every check should be done on backend too.

gdsoumya: applies for every place that updates/sets the password

gdsoumya (changes requested):

litmus-portal/authentication/api/handlers/user_handlers.go
```diff
@@ -0,0 +1,176 @@
...
+func UpdatePassword(service user.Service) gin.HandlerFunc {
+	return func(c *gin.Context) {
+		username := c.MustGet("username").(string)
+		uid := c.MustGet("uid").(string)
+		var adminUser entities.User
+		adminUser.UserName = username
+		var userPasswordRequest entities.UserPassword
+		err := c.BindJSON(&userPasswordRequest)
+		userPasswordRequest.Username = username
+		adminUser.ID, _ = primitive.ObjectIDFromHex(uid)
```
gdsoumya, line 120: Where is adminUser used in this function?

gdsoumya (changes requested):

litmus-portal/authentication/pkg/user/repository.go
```diff
@@ -0,0 +1,140 @@
...
+func (r repository) IsAdministrator(user *entities.User) error {
+	var result = entities.User{}
+	findOneErr := r.Collection.FindOne(context.TODO(), bson.M{
+		"_id":      user.ID,
+		"username": user.UserName,
+	}).Decode(&result)
+	if findOneErr != nil {
+		return findOneErr
+	}
+	if result.UserName != utils.AdminName || result.Role != entities.RoleAdmin {
```
gdsoumya, line 120: Unnecessary check for username, this prohibits creation of multiple admin accounts(which we can in the future) the role should only be the qualifying parameter.

gdsoumya (changes requested):

litmus-portal/authentication/api/handlers/user_handlers.go
DarthBenro008: is there any specific length of password length we need to abide by, or is it anything >= 2 characters?

gdsoumya: can you check the frontend, if it's not there we should add a standard one. Will have to discuss this if it's missing currently

DarthBenro008: @gdsoumya , I checked the frontend repo. We apparently are not doing any extensive checks and are just checking if the length is >= 0.

https://github.com/litmuschaos/litmus/blob/302c7051a03d4f6a75d0425eb0601150479d3ffa/litmus-portal/frontend/src/views/Settings/AccountsTab/AccountSettings/index.tsx#L62-L75

SarthakJain26: @DarthBenro008 Good to add standard checks for validation of username. For passwords let it be this way, we don't want to impose any hard checks on passwords as of now. cc: @gdsoumya

litmus-portal/authentication/api/handlers/user_handlers.go
gdsoumya: Why has this been resolved? What is the use of the `adminUser` variable being created?

gdsoumya (changes requested):

litmus-portal/authentication/api/handlers/user_handlers.go
gdsoumya: Not sure if not imposing strict rules is a good idea need to discuss this

DarthBenro008: @gdsoumya @SarthakJain26 So should we let this be as it is for the time being and create another issue related to this? As i think the conclusion of the discussion will require changes in both the places (the authentication back-end server as well as the frontend)

gdsoumya: We decided on adding a STRICT_PASSWORD_POLICY env var to the auth server, of this is set to true by the user the password policy will be used in the code else it will skip.

gdsoumya: By default it will be false. you will need to add the env var to the auth server manifest too

DarthBenro008: @gdsoumya have we decided on the password policy? if so, could you please share it?

gdsoumya: I think this is good for now :
1. at least 8 char long
2. at least 1 special character, 1 uppercase ,1 lowercase letter and 1 digit

Abstract the password check into a diff function so that it can be updated independently later.

DarthBenro008:
@gdsoumya @rajdas98, the sample deployment of the codebase can be found at: https://litmusportal-production-frontend-service-darthbenro008.cloud.okteto.net

Username: `admin`
Password: `litmus`

Strict policy check has been enabled for this deployment and can be tested by changing credentials.

litmus-portal/authentication/pkg/utils/sanitizers.go
```diff
@@ -1,7 +1,42 @@
...
+	"strings"
+)
 
+//SanitizeString trims the string input
 func SanitizeString(input string) string {
 	return strings.TrimSpace(input)
 }
+
+/*ValidateStrictPassword represents and checks for the following patterns:
+- Input more than 8 characters
```
gdsoumya, line 15: should be `at least 8 chars...`

gdsoumya (changes requested):

gdsoumya (approved):

PR #2953, opened 2021-06-29
Fixed bug in update/details api

SarthakJain26:

## Proposed changes

- UID will be taken from context for update user handler 
- Added condition to hash password before updating it in update/details API
- Updated bash script

## Types of changes

## Checklist

## Dependency

### DIFF
--- a/litmus-portal/authentication/api/handlers/user_handlers.go
+++ b/litmus-portal/authentication/api/handlers/user_handlers.go
@@ -46,15 +46,15 @@ func CreateUser(service user.Service) gin.HandlerFunc {
 }
 func UpdateUser(service user.Service) gin.HandlerFunc {
 	return func(c *gin.Context) {
-		uid := c.MustGet("uid").(string)
 		var userRequest entities.User
-		userRequest.ID = uid
 		err := c.BindJSON(&userRequest)
 		if err != nil {
 			log.Warn(err)
 			c.JSON(utils.ErrorStatusCodes[utils.ErrInvalidRequest], presenter.CreateErrorResponse(utils.ErrInvalidRequest))
 			return
 		}
+		uid := c.MustGet("uid").(string)
+		userRequest.ID = uid
 		userResponse, err := service.UpdateUser(&userRequest)
 		if err != nil {
 			c.JSON(utils.ErrorStatusCodes[utils.ErrServerError], presenter.CreateErrorResponse(utils.ErrServerError))
@@ -111,15 +111,15 @@ func LoginUser(service user.Service) gin.HandlerFunc {
 
 func UpdatePassword(service user.Service) gin.HandlerFunc {
 	return func(c *gin.Context) {
-		username := c.MustGet("username").(string)
 		var userPasswordRequest entities.UserPassword
 		err := c.BindJSON(&userPasswordRequest)
-		userPasswordRequest.Username = username
 		if err != nil {
 			log.Warn(err)
 			c.JSON(utils.ErrorStatusCodes[utils.ErrInvalidRequest], presenter.CreateErrorResponse(utils.ErrInvalidRequest))
 			return
 		}
+		username := c.MustGet("username").(string)
+		userPasswordRequest.Username = username
 		if utils.StrictPasswordPolicy {
 			err := utils.ValidateStrictPassword(userPasswordRequest.NewPassword)
 			if err != nil {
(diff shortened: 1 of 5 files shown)


gdsoumya (approved):

litmus-portal/authentication/pkg/user/repository.go
```diff
@@ -86,8 +89,15 @@ func (r repository) CreateUser(user *entities.User) (*entities.User, error) {
...
+	if user.Password != "" {
+		hashedPassword, err := bcrypt.GenerateFromPassword([]byte(user.Password), utils.PasswordEncryptionCost)
+		if err != nil {
+			return nil, err
+		}
+		user.Password = string(hashedPassword)
+	}
 	data, _ := toDoc(user)
-	_, err := r.Collection.UpdateOne(context.Background(), bson.M{"_id": user.ID}, bson.M{"$set": data})
+	_, err := r.Collection.UpdateOne(context.Background(), bson.M{"username": user.UserName}, bson.M{"$set": data})
```
DarthBenro008, line 100: Updating via username would be the right approach, but I think if we are migrating to updaing via username, we should get the username parsed from `JWT parsing` and not from the request body sent by frontend(client), as anyone can mimic a request with the username and update the details.

gdsoumya: Why do we need to use username instead of id @SarthakJain26 ?

SarthakJain26: @gdsoumya  ID and username both are unique, I updated it to username because on initial login where user is asked to update his/her password, we are sending username from the frontend. 
It was not giving any error because ID and username both have `omitempty` tags, which I think should be removed. cc: @DarthBenro008

gdsoumya: I think the jwt token will have the id right? we are getting the jwt token in the backend or are we not sending the token in this call?

SarthakJain26: Yes we are getting token. I think `JWT parsing` is good for both username as well as ID

gdsoumya: I think user id would be better then.

rajdas98: >I think user id would be better then.
then frontend needs to change the params Right?

gdsoumya: No, the data is being taken from jwt right so it should always be there @SarthakJain26 ?

litmus-portal/authentication/pkg/user/repository.go
```diff
@@ -16,7 +17,7 @@ type Repository interface {
 	FindUser(user *entities.User) (*entities.User, error)
 	UpdatePassword(userPassword *entities.UserPassword, isAdminBeingReset bool) error
 	CreateUser(user *entities.User) (*entities.User, error)
-	UpdateUser(user *entities.User) (*entities.User, error)
+	UpdateUser(c *gin.Context, user *entities.User) (*entities.User, error)
```
DarthBenro008, line 20: Passing the `*gin.Context` to the repository breaks the "Clean Architecture" principle the server follows as the repository shouldn't be aware of what the API framework does, in order to use the `uid` that is parsed from `JWT Token`, please call it directly at the handler, assign it to the user struct/password struct that is being passed to the repository. 

You can refer to the below snippet of code that achieves the same without passing context deeper into core logic.
 
https://github.com/litmuschaos/litmus/blob/b7243c9979d2e166fe472994c20e8fc4a7bc1d1d/litmus-portal/authentication/api/handlers/user_handlers.go#L112-L115

DarthBenro008 (changes requested):
Hey @SarthakJain26 , left a comment, PTAL.

litmus-portal/authentication/pkg/user/repository.go
SarthakJain26: > No, the data is being taken from jwt right so it should already be there @SarthakJain26 ?

Yes @gdsoumya

litmus-portal/authentication/api/handlers/user_handlers.go
```diff
@@ -46,7 +46,9 @@ func CreateUser(service user.Service) gin.HandlerFunc {
 }
 func UpdateUser(service user.Service) gin.HandlerFunc {
 	return func(c *gin.Context) {
+		uid := c.MustGet("uid").(string)
 		var userRequest entities.User
+		userRequest.ID = uid
```
DarthBenro008, line 51: I think the assignment can be done after binding the request body, incase if anything goes wrong.

DarthBenro008 (commented):
Left a trivial comment, everything else LGTM!

DarthBenro008 (approved):

PR #2964, opened 2021-07-05
Disable and Enable user from auth

SarthakJain26:

## Proposed changes

- `/updatestatus` API allows admin users to enable or disable a user's state.
- Disabled user will not be allowed to login. 
- Separated password validation from find user utility.

## Types of changes

## Checklist

## Dependency

- Username was considered in `/updatestatus` API because it uses findUser utility function which accepts username as input. 
findUser is used at the time of login, therefore it accepts username instead of uid.

### DIFF
--- a/litmus-portal/authentication/pkg/utils/errors.go
+++ b/litmus-portal/authentication/pkg/utils/errors.go
@@ -15,8 +15,8 @@ var (
 	ErrUserNotFound                  AppError = errors.New("user does not exists")
 	ErrWrongPassword                 AppError = errors.New("password doesn't match")
 	ErrUpdatingAdmin                 AppError = errors.New("cannot remove admin")
-	ErrUserRemoved                   AppError = errors.New("your account has been disabled")
-	ErrUserAlreadyRemoved            AppError = errors.New("user already removed")
+	ErrUserDeactivated               AppError = errors.New("your account has been deactivated")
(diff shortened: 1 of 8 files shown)


rajdas98 (approved):

litmus-portal/authentication/pkg/utils/errors.go
```diff
@@ -2,34 +2,43 @@ package utils
...
 	ErrInvalidRequest                AppError = errors.New("invalid_request")
 	ErrStrictPasswordPolicyViolation AppError = errors.New("password_policy_violation")
-	ErrUnauthorised                  AppError = errors.New("unauthorised")
+	ErrUnauthorized                  AppError = errors.New("unauthorized")
 	ErrUserExists                    AppError = errors.New("user_exists")
+	ErrUserNotFound                  AppError = errors.New("user does not exists")
+	ErrWrongPassword                 AppError = errors.New("password doesn't match")
+	ErrUpdatingAdmin                 AppError = errors.New("cannot remove admin")
+	ErrUserRemoved                   AppError = errors.New("your account has been disabled")
+	ErrUserAlreadyRemoved            AppError = errors.New("user already removed")
```
gdsoumya, line 19: The terminology should remain same, either it's removed or disabled. If the user can be re-enabled then it should be disabled.

gdsoumya (changes requested):

gdsoumya (approved):

PR #3098, opened 2021-08-06
Added checks to prevent DB inconsistency

SarthakJain26:

## Proposed changes

- Added a new route `/GetUser/:uid` that fetches the user details based on `uid`.
- Added checks to prevent Database inconsistencies due to caching. 

## Types of changes

## Checklist

## Dependency

### DIFF
--- a/litmus-portal/frontend/src/pages/GetStartedPage/index.tsx
+++ b/litmus-portal/frontend/src/pages/GetStartedPage/index.tsx
@@ -19,6 +19,7 @@ import {
   getUserId,
   getUsername,
   getUserRole,
+  logout,
 } from '../../utils/auth';
 import { validateConfirmPassword } from '../../utils/validate';
 import useStyles from './styles';
@@ -52,6 +53,31 @@ const GetStarted: React.FC = () => {
   const username = getUsername();
 
   const [loading, setIsLoading] = useState<boolean>(false);
+
+  // Checking if token is valid or not by finding the uid in database
+  const ValidateUser = () => {
+    fetch(`${config.auth.url}/getUser/${getUserId()}`, {
+      method: 'GET',
+      headers: {
+        'Content-Type': 'application/json',
+        Authorization: `Bearer ${getToken()}`,
+      },
+    })
+      .then((response) => response.json())
+      .then((data) => {
+        if ('error' in data) {
+          console.error('User not found');
+          logout();
+        }
+      })
+      .catch((err) => {
+        console.error(err);
+        logout();
+      });
+    return true;
+  };
+
+  ValidateUser();
 
   // Mutation to create project for a user
   const [CreateProject] = useMutation<Project>(CREATE_PROJECT, {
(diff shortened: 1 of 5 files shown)


gdsoumya (approved):

litmus-portal/frontend/src/pages/GetStartedPage/index.tsx
```diff
@@ -53,6 +54,31 @@ const GetStarted: React.FC = () => {
...
+      method: 'GET',
+      headers: {
+        'Content-Type': 'application/json',
+        Authorization: `Bearer ${getToken()}`,
+      },
+    })
+      .then((response) => response.json())
+      .then((data) => {
+        if ('error' in data) {
+          console.error('User not found');
```
Jonsy13, line 69: Can we display the error here also?

litmus-portal/frontend/src/pages/GetStartedPage/index.tsx
```diff
@@ -53,6 +54,31 @@ const GetStarted: React.FC = () => {
...
+        }
+      })
+      .catch((err) => {
+        console.error(err);
+        logout();
+      });
+    return true;
+  };
+
+  ValidateUser();
```
Jonsy13, line 80: Do we need this function call here? And if needed here, then it shouldn't be needed in onClick() of buttons?

litmus-portal/frontend/src/pages/GetStartedPage/index.tsx
```diff
@@ -190,8 +216,10 @@ const GetStarted: React.FC = () => {
                   title="Skip for now"
                   variant="highlight"
                   onClick={() => {
-                    setIsLoading(true);
-                    getUserInfo();
+                    if (ValidateUser()) {
```
Jonsy13, line 219: PTAL on above comment. Also If it is needed here then, it should be added in onClick() of Submit button.

Jonsy13 (changes requested):

litmus-portal/frontend/src/pages/GetStartedPage/index.tsx
SarthakJain26: We can display it, but if the UI encounters an error, then the cookie is cleared and page is refreshed and therefore everything from the console is cleared. It's quite difficult to even see that error

litmus-portal/frontend/src/pages/GetStartedPage/index.tsx
SarthakJain26: Yes, as soon as someone is redirected to /getStarted route, we check if the token details are valid. Also, whenever user enters a value in input field, the state of page changes and this function gets called. Therefore even if someone is already present on this page and tries to reset the password, then this will not allow them to do the changes. 
And since this function gets called on each input in the input field, therefore it is not required on onCompleted.

litmus-portal/frontend/src/pages/GetStartedPage/index.tsx
SarthakJain26: State changes on every input in the input field. But if a user directly clicks on skip button, then we should have a check to prevent any other API calls. That's why it's added here

Jonsy13 (approved):

arkajyotiMukherjee (approved):

Issue #3158, opened 2021-08-19
User details in My-accounts don't persist


Saranya-jena:
**What happened**:
The user details - name and email don't persist in their respective text fields.

**What you expected to happen**:
The user details should be displayed in the text fields as default values which can later be edit by the user.

**How to reproduce it (as minimally and precisely as possible)**:
This can be viewed in My-Account tab in Settings section.

**Anything else we need to know?**:
![Screenshot from 2021-08-19 22-52-30](https://user-images.githubusercontent.com/28081206/130114930-90c76701-43a1-4efb-a650-1698830a4931.png)

sayantani11:
Can I work on this?

Saranya-jena:
> Can I work on this?

Sure! Assigning it to you.

sayantani11:
@Saranya-jena Can you tell me how should I start?

Saranya-jena:
Hi @sayantani11 you can start with taking a look at the implementation of `My-Accounts` file:  https://github.com/litmuschaos/litmus/tree/master/litmus-portal/frontend/src/views/Settings/AccountsTab 

You can do the dev setup by referring to this doc : https://github.com/litmuschaos/litmus/wiki/Litmus-Portal-Development-Guide

sayantani11:
Sorry but I don't think I'll be able to continue with the issue. Sorry.
/unassign

duttaANI:
I would like to share my approach for future reference :

I am able to solve this issue partially by editing the following file -

**litmus/litmus-portal/frontend/src/views/Settings/AccountsTab/PersonalDetails/index.tsx**

Partially because the previously set values would be displayed on their respective fields (full name , email ) . However while editing the values , the user will not be able to backspace on the first letter . ( that is if user wants to edit it the following steps are required-

enter a new letter at index 0
Then delete the letter at index 1
 )

**Possible explanation** - empty string or "" is a falsey in typescript . However fullName and email fields as defined in the interface PersonaData are both type string .
Hence while handling changes the handle wont accept a false datatype or empty string  .


some links for reference -

[stack overflow](https://stackoverflow.com/questions/63354408/unable-to-delete-value-character-from-input-reactjs)

(https://github.com/redux-form/redux-form/issues/2537)[https://github.com/redux-form/redux-form/issues/2537]


**Modified code -**

`
import { useMutation, useQuery } from '@apollo/client';
import { Typography } from '@material-ui/core';
[...]

PR #3210, opened 2021-09-19
Feat: Added feature to display previos values of users fullname and e…

duttaANI:
…mail set in accounts tab

To discuss-
1) box at top right doesnt load new email when changed immediately

2) both values must be set and then save button gets activated

## Proposed changes

Summarize your changes here to communicate with the maintainers and make sure to put the link of that issue:

[issue: ](https://github.com/litmuschaos/litmus/issues/3158)

Changes are made to display full name and user email 

To handle empty string , the handleNameChange and handleEmailChange will strip "space" automatically while user updates values

## Types of changes

## Checklist

## Dependency

To enable save button both values i.e. full name and email must be set . Previously it was limited to full name only

### DIFF
--- a/litmus-portal/frontend/src/components/Header/ProfileDropDown.tsx
+++ b/litmus-portal/frontend/src/components/Header/ProfileDropDown.tsx
@@ -10,12 +10,15 @@ import { ButtonFilled, Icon, TextButton } from 'litmus-ui';
 import React, { useState } from 'react';
 import { useTranslation } from 'react-i18next';
 import { Link } from 'react-router-dom';
-import { GET_USER_INFO } from '../../graphql';
-import { CurrentUserDetails } from '../../models/graphql/user';
+import { GET_USER } from '../../graphql';
+import {
+  CurrentUserDetails,
+  CurrentUserDedtailsVars,
+} from '../../models/graphql/user';
 import useActions from '../../redux/actions';
 import * as TabActions from '../../redux/actions/tabs';
 import { history } from '../../redux/configureStore';
-import { getUserEmail, getUsername, logout } from '../../utils/auth';
+import { getUsername, logout } from '../../utils/auth';
 import { getProjectID, getProjectRole } from '../../utils/getSearchParams';
 import { userInitials } from '../../utils/userInitials';
 import useStyles from './styles';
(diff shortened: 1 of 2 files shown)


Saranya-jena (changes requested):
Can you remove all the unused comments?

Saranya-jena:
Is your approach partially solving the issue as per your comment: https://github.com/litmuschaos/litmus/issues/3158#issuecomment-922364793 ?

duttaANI:
I have solved the issue of displaying old full name and old email values set by the user .

After this I encountered a problem that user was not able to backspace the first letter which I had mentioned [here](https://github.com/litmuschaos/litmus/issues/3158#issuecomment-922364793).
However I solved this problem by replacing empty string by a "space" and  trimming the "space" from beginning of the string (since a condition is there for name to be valid it should not begin with a space) .

Now user would be able to edit both values without any issue.

Two extra points I would like to share with the developers  -

- First one is related to save button being activated.

code before this PR

```
<ButtonFilled
              disabled={
                !(
                  personaData.fullName.length &&
                  !loading
                )
              }
              onClick={handleSubmit}
            >
```

The issue with above code is that if the user chooses to only edit his email the save button still remains disabled .

To fix this I changed it to following -

```
<ButtonFilled
              disabled={
                !(
                  personaData.email.length &&
                  personaData.fullName.length &&
                  !loading
                )
              }
              onClick={handleSubmit}
            >
```

The catch here is that the user would have to click on both input fields to get the save button activated.

- Second - It seems that there is an issue with top right box , it doesn't update email value after user changes it . However if the user logs in again it would update


litmus

Saranya-jena:
I think the email is being extracted from JWT token `getUserEmail()` . In that case, it will only get updated when a new token is generated(after login). So instead of taking it from token , can you try using this query: `GET_USER_INFO`(Graphql endpoint) or `/getUser/:uid` (Rest endpoint) to fetch the user details and set the updated email?

duttaANI:
> Is your approach partially solving the issue as per your comment: [#3158 (comment)](https://github.com/litmuschaos/litmus/issues/3158#issuecomment-922364793) ?
> 
> The following approach should solve this issue completely:
> 
> ```
> const name: string = dataA?.getUser.name ?? ''; // TODO: Check if can be replaced with JWT based data.
>  const email: string = dataA?.getUser.email ?? '';
> ```
> 
> You can remove this section completely and assign empty strings as default values in `personaData `. Then add `onCompleted` in `getUser ` query where you'll be updating the `personaData` state with the latest data you got.
> 
> Also do add a `fetchPolicy` in the query and set it to `cache-and-network` . Can you try this approach?

Yes this approach solves the issue completely.

Saranya-jena (approved):

SarthakJain26 (approved):

Saranya-jena:
Thank you for the PR 🎉

duttaANI:
Thanks for guiding me towards correct approach.

PR #3242, opened 2021-10-01
area(litmusportal-auth): authentication backend to support dex redirects and callbacks

DarthBenro008:
Signed-off-by: DarthBenro008 <hkpdev008@gmail.com>

## Proposed changes

- Part of #2483 
- Enables [Dex OIDC](dexidp.io) Integration in litmusportal-authentication-backend
- New configurations related to OIDC and Dex
- Can enable and disable dex routed with the `DEX_ENABLED` environment variable

## Types of changes

## Checklist


litmus-portal/authentication/api/handlers/dex_auth_handler.go
```diff
@@ -0,0 +1,123 @@
...
+	"net/http"
+
+	"github.com/coreos/go-oidc/v3/oidc"
+	"github.com/gin-gonic/gin"
+	log "github.com/sirupsen/logrus"
+	"golang.org/x/oauth2"
+)
+
+var (
+	state    = "holderState"
```
gdsoumya, line 18: This has the same issues as the github/google oauth. The state can be overridden by another simultaneous call to the dex, so verification will not work. Please update the state management to the same method used for google/github oauth

gdsoumya (changes requested):

litmus-portal/authentication/api/handlers/dex_auth_handler.go
DarthBenro008: Good catch! In order to manage state, we would probably have to use db or an in-memory array to maintain the same. I prefer an in-memory array, which should work.

> Please update the state management to the same method used for google/github oauth

Which method are you referring to? There is no Google/GitHub auth implementation, hence we have Dex that handles it for us.

gdsoumya: Okay, we did not add it to litmus. For the state solution, I suggest using jwt tokens, create a short-lived jwt token that can be verified in the callback. If the jwt token is valid then allow the login. Using jwt mean we don't need either a db or in-memory storage.

DarthBenro008: JWT wouldn't make sense here imo, cause we will be redirecting the client to Dex Server and DexServer will be creating the callback, we cannot modify the callback from DexServer to send additional headers and pass JWT. Hence the case would fail.

Issue #3245, opened 2021-10-01
Subscriber pod restarts when workflow is created with invalid manifest


vamsiammineni:
**What happened**:

- Fetched the workflow manifest for pod-delete experiment with the help of getPredefinedExperimentYAML Query 
-  Received manifest seems faulty at  this line **->** labels:\n    subject: \" 
   {{workflow.parameters.appNamespace}}_kube-proxy\". Which I have realized later 
- Converted the received manifest yaml string to json string 
- Created a workflow with the with manfiest(json string) with the mutation createChaosWorkflow. cronSyntax was set to run the 
   chaosworkflow immediately
- Chaos Workflow was created. But, the subscriber crashes as soon as the server tries to run the workflow. 

**What you expected to happen**:

- GetPredefinedExperimentYAML query should return a valid manifest 
- There should be validation at server side to reject the requests which does not conform to a valid schema
- Subscriber should not crash in such cases. It should be handled properly

**How to reproduce it (as minimally and precisely as possible)**:

- Fetch the pod-delete workflow manifest from the chaoshub with gepPredefinedExperimentYAML query
- convert it to valid json string 
- create a workflow using the json manifest string with createChaosWorkflow

subscriber pod crashing can be noticed at this point with the following log event in the subscriber

`time="2021-10-01T07:07:01Z" level=fatal msg="error on processing request" error="error performing cluster operation: Workflow.argoproj.io \"some-name-6\" is invalid: metadata.labels: Invalid value: \"{{workflow.parameters.appNamespace}}_kube-proxy\": a valid label must be an empty string or consist of alphanumeric characters, '-', '_' or '.', and must start and end with an alphanumeric character (e.g. 'MyValue',  or 'my_value',  or '12345', regex used for validation is '(([A-Za-z0-9][-A-Za-z0-9_.]*)?[A-Za-z0-9])?')"`

**Anything else we need to know?**:

ksatchit:
cc: @rajdas98

rajdas98:
Hi @vamsiammineni, This is not a bug. this is a generic query (getPredefinedExperimentYAML) which is taking experiment name, hub name as input and searches in my hub repository (Default- [Myhub Repository](http://github.com/litmuschaos/chaos-charts))

[...]

PR #3402, opened 2021-12-20
Change fatal error to standard error for some logs

rajdas98:
Signed-off-by: rajdas98 <mail.rajdas@gmail.com>

## Proposed changes

fixes- https://github.com/litmuschaos/litmus/issues/3245#issuecomment-997934195

## Types of changes

## Checklist

## Dependency

### DIFF
--- a/litmus-portal/cluster-agents/subscriber/pkg/requests/webhook.go
+++ b/litmus-portal/cluster-agents/subscriber/pkg/requests/webhook.go
@@ -41,12 +41,12 @@ func ClusterConnect(clusterData map[string]string) {
 		}
 		data, err := json.Marshal(payload)
 		if err != nil {
-			logrus.WithError(err).Fatal("failed to marshal message")
+			logrus.WithError(err).Error("failed to marshal message")
 		}
 
 		err = c.WriteMessage(websocket.TextMessage, data)
 		if err != nil {
-			logrus.WithError(err).Fatal("failed to write message after init")
+			logrus.WithError(err).Error("failed to write message after init")
 			return
 		}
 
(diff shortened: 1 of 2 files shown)


Jonsy13 (approved):

gdsoumya:
Some of these need to be fatal errors, the logs were fatal deliberately because if the cluster connect has failed nothing else should be working on the subscriber. Please revert the changes @rajdas98

rajdas98:
> Some of these need to be fatal errors, the logs were fatal deliberately because if the cluster connect has failed nothing else should be working on the subscriber. Please revert the changes @rajdas98

@gdsoumya Cluster dail's error is in fatal- https://github.com/rajdas98/litmus/blob/e1edf723bc2d6c577cdbd7a5a8bd0513e05cd478/litmus-portal/cluster-agents/subscriber/pkg/requests/webhook.go#L33

gdsoumya:
everything inside the go function needs to be fatal, everything apart from the ReadMessage call in the loop can be errors but they need a continue statement after the log message.

rajdas98:
> @rajdas98 everything inside the go function needs to be fatal, everything apart from the ReadMessage call in the loop can be errors but they need a continue statement after the log message.

doesn't matter because we are not getting the payload data from the user. It's hardcoded.

rajdas98:
subscriber performs other activities such as sending events, logs through mutution. if we give fatal for WebSocket connection failure, then the other functionalities will stall.

gdsoumya:
That is necessary, this is how the subscriber was designed. The events and logs are pretty useless without the actual cluster connect, no one can ask for logs or k8s objects without the portal to cluster connection. The events will also cause issues because if the server is down or lost connectivity all the events sent will be lost, this was also a major reason for the restarts so that the subscriber when it connects fetches all the missed events and sends them to the portal.

rajdas98:
I know, I was thinking during the runtime of the subscriber, if the subscription has any issues, then we should run the other mutations because both mutation and subscription use different protocols. In some cases where WS protocol got blocked during the runtime and HTTP was running.

But I got your point, Will revert it soon.

PR #3430, opened 2022-01-14
(ChaosCenter): Fixed Upgrade Agent

Jonsy13:
Signed-off-by: Jonsy13 <vedant.shrotria@chaosnative.com>

## Proposed changes

## Types of changes

- [X] Bugfix (non-breaking change which fixes an issue)

## Checklist

- [X] I have read the [CONTRIBUTING](https://github.com/litmuschaos/litmus/blob/master/CONTRIBUTING.md) doc
- [X] I have signed the commit for DCO to be passed.

## Dependency

### DIFF
--- a/litmus-portal/upgrade-agents/control-plane/versions/upgrade-manager.go
+++ b/litmus-portal/upgrade-agents/control-plane/versions/upgrade-manager.go
@@ -109,8 +109,11 @@ func (m *UpgradeManager) Run() error {
 	versionIterator := m.PreviousVersion
 	// loop till the target version is reached
 	for versionIterator != m.TargetVersion {
-		if err := upgradePath[versionIterator].VersionManager.Run(); err != nil {
-			return fmt.Errorf("failed to upgrade to version %v, error : %w", versionIterator, err)
+		// Skipping schema upgrade, if version manager not available (Only version will be upgraded)
+		if upgradePath[versionIterator].VersionManager != nil {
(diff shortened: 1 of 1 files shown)


gdsoumya (approved):

rajdas98 (approved):

Jonsy13:
Hi @lamc , You can ignore that error that is because the version of controlplane in both the charts is same, only version of chart is different due to some changes in chart configs. Upgrade-agent is only concerned with upgrading controlplane from 2.4.0 to 2.5.0.
@gdsoumya can give more details.

lamc:
@Jonsy13 thank you for the quick response, however since I'm automating deployment through spinnaker the deployment fails while waiting for manifest to stabilize with error: `Job has reached the specified backoff limit` I looked at the template and default is 2.  How would you suggest proceeding? I will also wait for more details from @gdsoumya

gdsoumya:
I think it's necessary that the user is notified about the version being the same, you will only run upgrade-agent when you want an upgrade from an older version (it isn't run automatically every time) thus it is important for the user to be notified when a scenario like this occurs where the versions are already same as it is very different from what the user has assumed while running the upgrade-agent. A silent completion in such a case will make this undetectable.

lamc:
@gdsoumya @Jonsy13 thank you it the explanation! this makes sense and I agree that this error notification shouldn't be silenced. For my scenario I will just ignore the failure in my spinnaker pipeline when deploying this helm chart. Thank you so much!! :)

PR #3444, opened 2022-02-03
chore(chaos-center): Updated MyHub Rbacs and optimized filetype for fetching hub data

amityt:
Signed-off-by: Amit Kumar Das <amit@chaosnative.com>

## Proposed changes

This PR fixes the following issues:
- Fix the rbac issues with hub
- Updated Filetype as an `enum`

ref:  https://github.com/litmuschaos/litmus-e2e/issues/362

## Types of changes

## Checklist

## Dependency


litmus-portal/graphql-server/graph/schema.resolvers.go
```diff
@@ -123,7 +122,13 @@ func (r *mutationResolver) SaveMyHub(ctx context.Context, myhubInput model.Creat
...
 }
 
-func (r *mutationResolver) SyncHub(ctx context.Context, id string) ([]*model.MyHubStatus, error) {
+func (r *mutationResolver) SyncHub(ctx context.Context, id string, projectID string) ([]*model.MyHubStatus, error) {
+	err := authorization.ValidateRole(ctx, projectID,
+		authorization.MutationRbacRules[authorization.UpdateChaosWorkflow],
+		model.InvitationAccepted.String())
+	if err != nil {
+		return nil, err
+	}
```
gdsoumya, line 131: Will not work creates attack vector. I can send a random project id and an unrelated hub id, the project id should be derived from the hub id

litmus-portal/graphql-server/graph/schema.resolvers.go
```diff
@@ -163,7 +168,13 @@ func (r *mutationResolver) UpdateMyHub(ctx context.Context, myhubInput model.Upd
...
 }
 
-func (r *mutationResolver) DeleteMyHub(ctx context.Context, hubID string) (bool, error) {
+func (r *mutationResolver) DeleteMyHub(ctx context.Context, hubID string, projectID string) (bool, error) {
+	err := authorization.ValidateRole(ctx, projectID,
+		authorization.MutationRbacRules[authorization.DeleteMyHub],
+		model.InvitationAccepted.String())
+	if err != nil {
+		return false, err
+	}
```
gdsoumya, line 177: same attack vector

litmus-portal/graphql-server/graph/schema.resolvers.go
```diff
@@ -407,14 +418,32 @@ func (r *queryResolver) GetHubStatus(ctx context.Context, projectID string) ([]*
...
 	return myhub.GetPredefinedWorkflowList(hubName, projectID)
 }
 
 func (r *queryResolver) GetPredefinedExperimentYaml(ctx context.Context, experimentInput model.ExperimentInput) (string, error) {
+	err := authorization.ValidateRole(ctx, experimentInput.ProjectID,
+		authorization.MutationRbacRules[authorization.PredefinedWorkflowOperations],
+		model.InvitationAccepted.String())
+	if err != nil {
+		return "", err
+	}
```
gdsoumya, line 446: Make sure the ProjectIDs passed here are actually used to get the required data else, if they are just redundant information the same attack vector exists here too.

gdsoumya (changes requested):

litmus-portal/graphql-server/graph/schema.resolvers.go
gdsoumya: A simple solution would be to query the hub with the hub id + project id, this would mean if the user sends unrelated hub and project ids db will not return the hub data.

amityt:
Thanks @gdsoumya  for the review 🚀 
I have made the required changes, PTAL!

litmus-portal/graphql-server/graph/schema.resolvers.go
amityt: @gdsoumya  For the above 2 cases, the Project ID is being used to fetch the chaos-charts data from the hub directories. I believe it wont cause any issue here. Have updated the code for other cases.

gdsoumya: Awesome thanks.

litmus-portal/graphql-server/pkg/myhub/myhub.go
```diff
@@ -292,9 +296,15 @@ func GetYAMLData(ctx context.Context, experimentInput model.ExperimentInput) (st
...
+// GetPredefinedExperimentYAMLData is responsible for sending the workflow.yaml for a given pre-defined workflow.
 func GetPredefinedExperimentYAMLData(ctx context.Context, experimentInput model.ExperimentInput) (string, error) {
-	YAMLPath := handler.GetPredefinedExperimentManifest(ctx, experimentInput)
+	var YAMLPath string
+	if strings.ToLower(experimentInput.FileType.String()) != "workflow" {
+		return "", errors.New("invalid file type")
+	}
+	if strings.ToLower(experimentInput.ChartName) == "predefined" && strings.ToLower(experimentInput.FileType.String()) == "workflow" {
+		YAMLPath = handler.GetPredefinedExperimentManifest(ctx, experimentInput)
+	}
```
gdsoumya, line 307: The file type checks isn't good enough, if the file type can change how the data is stored/operated on then it'll be better to check the actual k8s object and verify the `Kind` of the resource being sent.

gdsoumya (changes requested):

litmus-portal/graphql-server/pkg/myhub/myhub.go
amityt: @gdsoumya FileType is basically used to distinguish between different files from the chaos-charts repository and generate the hub path for the same. Not sure if we need another check to verify the `Kind` here.

gdsoumya (approved):

Jonsy13 (approved):

Issue #3597, opened 2022-05-13
Fix agent connectivity for control plane exposed via non nginx-controller based ingress 


ksatchit:
Today, litmus supports the successful resolution of server address (SERVER_ADDR on the agent-config configmap) in cases where the nginx ingress controller is used (ref: https://docs.litmuschaos.io/docs/user-guides/setup-with-ingress/#install-chaoscenter-with-ingress). However, this same capability is not extended to other ingress types (AWS ALB, Trafeik, istio gateway etc.,). 

This issue tracks a need for a generic solution that would enable successful setup/connectivity of agent regardless of the ingress type used.

ksatchit:
A proposed solution is to use a tunable (env) on the chaos server deployment (say, SERVER_URL) that can be injected into the config as part of agent setup. Refer to conversations within the community regarding this, here: https://kubernetes.slack.com/archives/CNXNB0ZTN/p1651590192986229

PR #3632, opened 2022-05-26
Adding CHAOS_CENTER_UI_ENDPOINT available for cluster and namespace scope

rajdas66:
## Proposed changes

Issue- https://github.com/litmuschaos/litmus/issues/3597

Changes:
* Adding CHAOS_CENTER_UI_ENDPOINT available for cluster and namespace scope
* Renaming PORTAL_SCOPE to CHAOS_CENTER_SCOPE

## Types of changes

## Checklist

## Dependency


litmus-portal/graphql-server/pkg/handlers/file_handler.go
```diff
@@ -53,45 +60,48 @@ func FileHandler(w http.ResponseWriter, r *http.Request) {
...
 	}
 
-	scope := os.Getenv("PORTAL_SCOPE")
-	if scope == clusterScope {
+	if scope == namespaceScope && endpoint == "" {
+		return nil, 500, errors.New("CHAOS_CENTER_UI_ENDPOINT env is empty for namespace scope installation")
+	}
+
+	if endpoint != "" {
+		subscriberConfiguration.GQLServerURI = endpoint + "/query"
```
Jonsy13, line 78: If user is providing UI url, then appending `/query` might not work, right? According to nginx, it will be available `/ws/query`, `/query` will only work for NodePort.

gdsoumya (changes requested):

litmus-portal/graphql-server/pkg/handlers/file_handler.go
gdsoumya: yes it needs to be /ws/query as we are asking for the UI endpoint, even for loadbalancer or nodeport it needs to be /ws/query because we are always asking for UI endpoint and not graphql server endpoint

litmus-portal/graphql-server/pkg/handlers/file_handler.go
```diff
@@ -53,45 +60,48 @@ func FileHandler(w http.ResponseWriter, r *http.Request) {
...
-	reqCluster, err := dbOperationsCluster.GetCluster(id)
+	reqCluster, err := dbOperationsCluster.GetCluster(clusterID)
 	if err != nil {
 		return nil, 500, err
 	}
 
-	scope := os.Getenv("PORTAL_SCOPE")
-	if scope == clusterScope {
+	if scope == namespaceScope && endpoint == "" {
+		return nil, 500, errors.New("CHAOS_CENTER_UI_ENDPOINT env is empty for namespace scope installation")
```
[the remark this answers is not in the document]
rajdas66: @gdsoumya Self agent will work as it is..

1. If `controlplane scope is namespace` then, `CHAOS_CENTER_UI_ENDPOINT is required`. (that check wasn't present earlier)
2. If `controlplane scope is cluster and CHAOS_CENTER_UI_ENDPOINT is not present`, then `it will retrieve the endpoint from the cluster(existing logic)`

So, as part of some discussions, we are making CHAOS_CENTER_UI_ENDPOINT available for both scopes.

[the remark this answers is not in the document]
gdsoumya: I understand that but how will the user know beforehand what their chaos center UI endpoint is ? If they are running nodeport or just port-forward it's difficult to determine at installation time.  The only option the user will have in such case is to deploy the chaos center and then figure the ip/url and re-apply with changed config because even if they just want to playaround with self-agent and no external agent in ns scope the self-agent installation will fail without the UI endpoint en var set.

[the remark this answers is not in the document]
rajdas66: for namespace mode- we already set the CHAOS_CENTER_UI_ENDPOINT to http://litmusportal-frontend-service:9091 as default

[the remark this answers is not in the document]
rajdas66: for cluster scope only we are not setting it.

gdsoumya (approved):

litmus-portal/graphql-server/pkg/handlers/file_handler.go
```diff
@@ -104,44 +120,48 @@ func GetManifest(token string) ([]byte, int, error) {
...
 	}
 
 	// Checking if cluster with given clusterID and accesskey is present
-	if reqCluster.AccessKey != key {
-		return nil, errors.New("Invalid access key")
+	if reqCluster.AccessKey != accessKey {
+		return nil, errors.New("Access Key is invalid")
+	}
+
+	if scope == namespaceScope && endpoint == "" {
```
Jonsy13, line 134: Can we also use `GetEndpoint` function here as well?

litmus-portal/graphql-server/pkg/handlers/file_handler.go
```diff
@@ -104,44 +120,48 @@ func GetManifest(token string) ([]byte, int, error) {
...
+		return nil, errors.New("CHAOS_CENTER_UI_ENDPOINT env is empty for namespace scope installation")
 	}
-	scope := os.Getenv("PORTAL_SCOPE")
-	if scope == clusterScope {
+
+	if endpoint != "" {
+		subscriberConfiguration.GQLServerURI = endpoint + "/query"
```
Jonsy13, line 139: This will also be `/ws/query`

rajdas66: Done

Jonsy13 (approved):

PR #3648, opened 2022-06-08
feat: Added UID to workflows instead of username

vanshBhatia-A4k9:
## Proposed changes

workflows now show UID instead of username as a unique identifier

## Types of changes

## Checklist

## Dependency


litmus-portal/graphql-server/pkg/cluster/handler/handler.go
```diff
@@ -277,7 +277,7 @@ func SendRequestToSubscriber(subscriberRequest clusterOps.SubscriberRequests, r
 			Namespace:    subscriberRequest.Namespace,
 			RequestType:  subscriberRequest.RequestType,
 			ExternalData: subscriberRequest.ExternalData,
-			Username:     subscriberRequest.Username,
+			UID:          subscriberRequest.Username,
```
Jonsy13, line 280: I thin, this should be UID
```suggestion
			UID:          subscriberRequest.UID,
```

litmus-portal/graphql-server/pkg/chaos-workflow/ops/operations.go
```diff
@@ -248,7 +248,7 @@ func SendWorkflowToSubscriber(workflow *model.ChaosWorkFlowRequest, username *st
 		ClusterID:    workflow.ClusterID,
 		Namespace:    workflowNamespace,
 		ExternalData: externalData,
-		Username:     username,
+		Username:     uid,
```
Jonsy13, line 251: This should be UID
```suggestion
		UID:     uid,
```

Jonsy13 (changes requested):

gdsoumya:
UIDs will probably solve the current issue but this also makes it less auditable unless we let the users/admins query accounts using uids. I don't remember having such UI currently but maybe I am wrong @vanshBhatia-A4k9

amityt:
Hi @brnck Thanks for reaching out! After some discussion with the team, we have decided to fix the issue from the DEX server itself, also we'll add some additional checks in the auth server to prevent any special characters in the username. It should be available by 2.13 as it is a larger change and will take some time to test the same.

@gdsoumya we'll be adding the user metadata that contains the username as well in the schema, but we'll only pass the userID in the manifest.

bbarin:
As @gdsoumya mentioned, changed to UUID will solve but breaks the auditing, which is very important. I would rather base64 the username to circumvent this issue.

Saranya-jena:
These changes have been added as part of 3.0

PR #3701, opened 2022-07-25
Adding delete check in all delete resource operations

imrajdas:
Signed-off-by: Raj Babu Das <mail.rajdas@gmail.com>

## Proposed changes

Currently, all delete APIs are responding success if the API is hit more than once, This change will restrict that.

## Types of changes

## Checklist

## Dependency

### DIFF
--- a/litmus-portal/graphql-server/pkg/chaos-workflow/handler/handler.go
+++ b/litmus-portal/graphql-server/pkg/chaos-workflow/handler/handler.go
@@ -81,6 +81,10 @@ func DeleteChaosWorkflow(ctx context.Context, projectID string, workflowID *stri
 		return false, err
 	}
 
+	if workflow.IsRemoved {
+		return false, errors.New("chaos workflow not found (ID: " + *workflowID + " )")
+	}
+
 	tkn := ctx.Value(authorization.AuthKey).(string)
 	username, err := authorization.GetUsername(tkn)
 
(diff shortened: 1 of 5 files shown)


litmus-portal/graphql-server/pkg/chaos-workflow/handler/handler.go
```diff
@@ -81,6 +81,10 @@ func DeleteChaosWorkflow(ctx context.Context, projectID string, workflowID *stri
 		return false, err
 	}
 
+	if workflow.IsRemoved {
```
SarthakJain26, line 84: Instead of this, you may apply filter in the query itself: `{"is_removed": true}`, and add a check for that.

litmus-portal/graphql-server/pkg/cluster/handler/handler.go
```diff
@@ -175,6 +180,7 @@ func DeleteClusters(ctx context.Context, projectID string, clusterIds []*string,
 	if err != nil {
 		return "", err
 	}
+
 	clusters, err := dbOperationsCluster.ListClusters(ctx, query)
```
SarthakJain26, line 184: This function should ideally not return any clusters, since we are updating the documents with `is_removed: true` and then we are querying the same documents with `is_removed: false`

litmus-portal/graphql-server/pkg/chaos-workflow/handler/handler.go
SarthakJain26, line 84: Instead of this, you may apply filter in the query itself: {"is_removed": true}, and add a check for that.

SarthakJain26 (changes requested):

PR #3780, opened 2022-09-29
feat(selectors): Adding multiple workloads and pod filtering

ispeakc0de:
Signed-off-by: Shubham Chaudhary <shubham.chaudhary@harness.io>

## Proposed changes

## Types of changes

## Checklist

## Dependency


amityt (approved):

Saranya-jena (approved):

litmus-portal/cluster-agents/subscriber/pkg/workloads/workloads.go
```diff
@@ -0,0 +1,59 @@
...
+			return nil, err
+		}
+		for k, v := range svc.Spec.Selector {
+			res, err := client.CoreV1().Pods(svc.Namespace).List(context.Background(), v1.ListOptions{LabelSelector: fmt.Sprintf("%s=%s", k, v)})
+			if err != nil {
+				return nil, err
+			}
+			for _, p := range res.Items {
+				pods = append(pods, p.Name)
+			}
```
gdsoumya, line 41: There could be possible overlaps for multiple selectors and as far as I remember services operate on  `and` operator for labels so all the labels should be present in the pod to be targeted by a service. In that case the pod listing should be done differently where all the labels are used.

gdsoumya (changes requested):

litmus-portal/cluster-agents/subscriber/pkg/workloads/workloads.go
```diff
@@ -0,0 +1,59 @@
...
+
+var managedByReplicaSet = func(ns, kind string, wld []string, client *kubernetes.Clientset) ([]string, string, error) {
+	rs, err := getReplicaSetsFromWorkload(ns, kind, wld, client)
+	return rs, ReplicaSet, err
+}
+
+var managedByReplicaController = func(ns, kind string, wld []string, client *kubernetes.Clientset) ([]string, string, error) {
+	rc, err := getReplicaControllersFromWorkload(ns, kind, wld, client)
+	return rc, ReplicationController, err
+}
```
gdsoumya, line 55: This code including the logic to fetch and match owners for each of replicaset and replica controller can be done in a single function if we use dynamic client, providing the proper gvr fopr each type we would get an `unstructured.Unstructured` response from there we can get the ownerref using `GetOwnerReferences` which is all we care about as far as I can see. That would make the code much more clean and reusable.

gdsoumya (changes requested):

SarthakJain26 (approved):

litmus-portal/cluster-agents/subscriber/pkg/workloads/workloads.go
ispeakc0de: good catch! I didn't notice that.
Yes, we can directly use it to list all pods with matching selectors with & operator
```
pods, err := client.CoreV1().Pods(svc.Namespace).List(context.Background(), v1.ListOptions{LabelSelector: svc.Spec.Selector})
```

litmus-portal/cluster-agents/subscriber/pkg/workloads/workloads.go
ispeakc0de: gvr  & dc can be used better if we traverse from parent to child. But we are deriving OwnerRef from child to parent `pod` --> `rs/rc/ds/sts` --> `dc/deploy/rollout`

But I will see if I can use it for sub-tasks

gdsoumya: I was thinking of something like this :
```
func getChildResourceFromParentresource(ns, kind string, wld []string, dynamicClient dynamic.Interface, childGVR schema.GroupVersionResource) ([]string, error) {

	resources, err := dynamicClient.Resource(childGVR).Namespace(ns).List(context.Background(), metav1.ListOptions{})
	if err != nil {
		return nil, err
	}

	var rs []string
	for _, r := range resources.Items {
		if checkOwnerRef(r.GetOwnerReferences(), wld, kind) {
			rs = append(rs, r.GetName())
		}
	}
	return rs, nil
}
```

Then you can just have one single function instead of having separate one for getting replicaset/replicacontroller/pods from their parent resources

gdsoumya (approved):

PR #3792, opened 2022-10-17
Adding SingleBranch:  true in the git clone options

imrajdas:
Reverts litmuschaos/litmus#3776

### DIFF
--- a/litmus-portal/graphql-server/pkg/myhub/ops/gitops.go
+++ b/litmus-portal/graphql-server/pkg/myhub/ops/gitops.go
@@ -90,15 +90,15 @@ func (c MyHubConfig) getChaosChartRepo() (string, error) {
 	ClonePath := GetClonePath(c)
 	os.RemoveAll(ClonePath)
 	_, err := git.PlainClone(ClonePath, false, &git.CloneOptions{
-		URL: c.RepositoryURL, 
-		Progress: os.Stdout,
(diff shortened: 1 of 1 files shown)


gdsoumya (changes requested):
We should only clone a single branch to reduce unnecessary data being downloaded.
This clone option should force only the required branch data to be cloned : 
```
SingleBranch:  true,
```

imrajdas:
> We should only clone a single branch to reduce unnecessary data being downloaded. This clone option should force only the required branch data to be cloned :
> 
> ```
> SingleBranch:  true,
> ```

@AmitKumarDas - Will it work with the git pull in go-git?

gdsoumya:
> > We should only clone a single branch to reduce unnecessary data being downloaded. This clone option should force only the required branch data to be cloned :
> > ```
> > SingleBranch:  true,
> > ```
> 
> @amityt - Will it work with the git pull in go-git?

It already works and is tested. I did the same with gitops feature ref: https://github.com/litmuschaos/litmus/blob/45693e9b31804a8d6dc3adb4020c57399fe16499/litmus-portal/graphql-server/pkg/gitops/gitops.go#L178

gdsoumya (approved):

gdsoumya:
@imrajdas LGTM, but lint check and dco is failing, please fix before merging

Jonsy13 (approved):

Issue #3931, opened 2023-03-22
Move to using interfaces for better mocking, abstraction and decoupling


S-ayanide:
Currently, our Golang codebase heavily relies on concrete types for implementing functionality. While this approach has its benefits, it can lead to tight coupling between modules and make our codebase less modular and harder to maintain in the long run.

To address this, we propose that we start using interfaces in our codebase. Interfaces provide a way to define a set of behaviors that a type must implement, without specifying the implementation details. This allows us to achieve better abstraction and decoupling, making our code more modular and easier to test.

An example template on how we can move forward with this -
https://gist.github.com/S-ayanide/94654714b9f1a7f5da3a7cdf38396427

namkyu1999:
As an LFX mentee, I can handle this issue!

PR #3933, opened 2023-03-25
Move to using interface for better mocking at chaoshub package

namkyu1999:
## Proposed changes

Currently, our Golang codebase heavily relies on concrete types for implementing functionality. It causes hard to make unit tests. So I propose an interface pattern for loose coupling like this PR. For this PR, I only targeted the chaoshub package. If this approach is correct. I will refactor all packages to the interface.

And last, The below variable will be deleted after all packages are refactored to the interface.
https://github.com/litmuschaos/litmus/blob/601da40aae7c3bae19fd5dc8fd4b133021e62f3d/litmus-portal/graphql-server/pkg/database/mongodb/operations.go#L33

Issue No. #3931 

## Types of changes

## Checklist

## Dependency


S-ayanide (approved):
Clean code 🚀 Hub looks pretty refactored to me 🏆

litmus-portal/graphql-server/graph/resolver.go
```diff
@@ -8,17 +8,21 @@ import (
...
+	"github.com/litmuschaos/litmus/litmus-portal/graphql-server/pkg/chaoshub"
 )
 
 // This file will not be regenerated automatically.
 //
 // It serves as dependency injection for your app, add any dependencies you require here.
 
-type Resolver struct{}
+type Resolver struct {
+	ChaosHubService chaoshub.Service
```
gdsoumya, line 19: This field does not need to be exported
```suggestion
	chaosHubService chaoshub.Service
```

litmus-portal/graphql-server/graph/model/modes_factory.go
```diff
@@ -0,0 +1,16 @@
...
+		RepoURL:       chaosHub.RepoURL,
+		HubName:       chaosHub.HubName,
+		IsPrivate:     chaosHub.IsPrivate,
+		UserName:      chaosHub.UserName,
+		Password:      chaosHub.Password,
+		AuthType:      chaosHub.AuthType,
+		Token:         chaosHub.Token,
+		SSHPrivateKey: chaosHub.SSHPrivateKey,
+	}
+}
```
gdsoumya, line 16: Might be a good idea to move this to the `pkg/chaoshub`, the `graph/model` has auto-generated code and can be over-written in the future by mistake. This function is also specific to `pkg/chaoshub` so it makes more sense to have it in there

litmus-portal/graphql-server/pkg/chaoshub/service.go
```diff
@@ -15,48 +15,63 @@ import (
...
+}
 
-	IsExist, err := IsChaosHubAvailable(ctx, chaosHub.HubName, chaosHub.ProjectID)
-	if err != nil {
-		return nil, err
-	}
-	if IsExist == true {
-		return nil, errors.New("HubName Already exists")
+type chaosHubService struct {
+	ChaosHubOperation dbSchemaChaosHub.Operator
```
gdsoumya, line 50: same here no need to export
```suggestion
	chaosHubOperation dbSchemaChaosHub.Operator
```

litmus-portal/graphql-server/pkg/chaoshub/service.go
```diff
@@ -15,48 +15,63 @@ import (
...
-		RepoURL:       chaosHub.RepoURL,
-		HubName:       chaosHub.HubName,
-		IsPrivate:     chaosHub.IsPrivate,
-		UserName:      chaosHub.UserName,
-		Password:      chaosHub.Password,
-		AuthType:      chaosHub.AuthType,
-		Token:         chaosHub.Token,
-		SSHPrivateKey: chaosHub.SSHPrivateKey,
+	// go routine for syncing hubs for all users
+	go chaosHubService.RecurringHubSync()
```
gdsoumya, line 59: Possible Bug, this function is called in 2 places :
1. GQL Resolver
2. ProjectInitializer function

According to this implementation, the only place the go routine should be started is when the resolver is set but because ProjectIntializer also calls this function every time a new project is created duplicate goroutines will start which will lead to excessive memory consumption.

The best thing to do is to remove this goroutine call from here and put it directly in the `main()` like it was before. Having a goroutine start inside some obscure function call is probably not the best idea especially when there should only be 1 such routine in a given instance of the server.

gdsoumya (changes requested):

litmus-portal/graphql-server/pkg/database/mongodb/chaoshub/operations.go
```diff
@@ -9,9 +9,25 @@ import (
...
+	UpdateChaosHub(ctx context.Context, query bson.D, update bson.D) error
+}
+
+type chaosHubOperator struct {
+	operator mongodb.MongoOperator
+}
+
+func NewOperator() Operator {
+	return &chaosHubOperator{operator: &mongodb.MongoOperations{}}
+}
```
gdsoumya, line 26: We probably don't need to convert this to an interface because `mongodb.Operator` which was previously used in the functions is already an interface type (`MongoOperator`). If this is for mocking in tests imo just creating an in memory db mock that implements `mongodb.MongoOperator` would suffice instead of having to mock every single db wrapper that we have for each service. That will be much easier and faster imo. 

Though I agree that we should not be using a global shared variable `mongodb.Operator` so the best change would be to just create a strict type for each service (in this case ChaosHubOperator) and in the initializer, we take the db implementation needed something like this :

```
func NewChaosHubOperator(dbOperator mongodb.MongoOperator) ChaosHubOperator {
	return &ChaosHubOperator{operator: dbOperator}
}
```
That way we can substitute the db instance to mock db in our tests without having to write mocks for each individual service wrapper. WDYT?

gdsoumya (changes requested):

litmus-portal/graphql-server/graph/resolver.go
amityt: Thinking aloud, this `Resolver` should be used for dependency injection,  we can also initialize the mongo client here and use the same mongo client in all our handlers. Was checking the official documentation and found this here: https://gqlgen.com/getting-started/

`The graph/resolver.go file is where we declare our app’s dependencies, like our database. It gets initialized once in server.go when we create the graph.`

`type Resolver struct{
	todos []*model.Todo
}`

amityt: For the same approach here as well: https://hasura.io/blog/building-a-graphql-api-with-golang-postgres-and-hasura/

amityt (changes requested):

litmus-portal/graphql-server/graph/resolver.go
gdsoumya: It would be unnecessary as we are already storing the individual service instance(for eg. chaosHubService). The service object would have the db instance so no point in storing it separately as our resolvers don't directly use the db instance. We can do initialization like :

```
chaoshub.NewService(dbInstance)
```

If we want to parameterize the dbInstance being passed which is a good idea.

amityt: Yes, we can do something like that ^^

litmus-portal/graphql-server/graph/resolver.go
```diff
@@ -9,19 +9,21 @@ import (
...
 //
 // It serves as dependency injection for your app, add any dependencies you require here.
 
 type Resolver struct {
-	ChaosHubService chaoshub.Service
+	chaosHubService chaoshub.Service
 }
 
 func NewConfig() generated.Config {
+	dbOperator := &mongodb.MongoOperations{}
```
gdsoumya, line 24: This can be passed from the main() itself, we can share the same db instance and use it across all the services and go routines. We are already creating a db instance in the main for the recurring hub sync let's reuse that. In the future when all the services are refactored we will just have 1 single db instance passed around to all of them too.

gdsoumya (changes requested):
Rest looks good

litmus-portal/graphql-server/pkg/projects/project_handler.go
```diff
@@ -55,8 +56,9 @@ func ProjectInitializer(ctx context.Context, projectID string, role string) erro
 	}
 
 	log.Print("Cloning https://github.com/litmuschaos/chaos-charts")
+
 	//TODO: Remove goroutine after adding hub optimisations
-	go chaoshub.AddChaosHub(context.Background(), defaultHub)
+	go chaoshub.NewService(&mongodb.MongoOperations{}).AddChaosHub(context.Background(), defaultHub)
```
namkyu1999, line 61: this line also creates new service instance because service logic is initialized in graph package.

litmus-portal/graphql-server/graph/resolver.go
namkyu1999: So, Instead of using a global variable like the below, Parameterize the db instance to `NewConfig(dbInstance)` and inject them in the main function. Is it right? @gdsoumya 


https://github.com/litmuschaos/litmus/blob/601da40aae7c3bae19fd5dc8fd4b133021e62f3d/litmus-portal/graphql-server/pkg/database/mongodb/init.go#L57

gdsoumya: Yup, we should move away from global variables and multiple separate initializations as much as we can.

namkyu1999: Okay I will figure it out 🚀

imrajdas (approved):

PR #3937, opened 2023-03-31
Move to using interface for better mocking to graphql server

namkyu1999:
## Proposed changes
Following up on the last PR, I've converted all packages to the interface pattern for better mocking in graphql server. I made this PR based on [this layered architecture](https://github.com/bxcodec/go-clean-arch). 

Issue No. #3931 
Prev PR. #3933 

## Types of changes

## Checklist

## Dependency


namkyu1999:
I fixed all codes using interface. So no more global variable `mongo operator` now 😄

litmus-portal/graphql-server/server.go
```diff
@@ -80,10 +89,8 @@ func main() {
 	mongoClient := mongodb.Initialize(client)
 
 	var mongodbOperator mongodb.MongoOperator = mongodb.NewMongoOperations(mongoClient)
-	// TODO: remove this when all packages shift to interface pattern
-	mongodb.Operator = mongodbOperator
```
gdsoumya, line 84: There are some global variables still there in the database package I believe we need to remove those too right?

gdsoumya (changes requested):

litmus-portal/graphql-server/server.go
namkyu1999: Move some variables to local in this commit. If any variables needed to be fixed, Please tell me! [fix: move variables global to local](https://github.com/litmuschaos/litmus/pull/3937/commits/ab3b0567e7e9af616849fc92181ff347d5bf1fe9)

namkyu1999:
ab
A is current Changes, and B is my proposals.
I made this PR based on [this layered architecture](https://github.com/bxcodec/go-clean-arch). </br>
For example in the chaoshub package, As of now A, When I make a unit test in the ChaosHubService(service layer), I cannot create ChaosHubOperator Mock(repository layer) because ChaosHubOperator is not an interface but a struct. So, For better testing, I propose B. I want to know your opinion. @imrajdas @gdsoumya @amityt @S-ayanide

gdsoumya:
@namkyu1999 currently the A approach needs you to mock the dboperator instead of the service specific operator so you mock every db method that exists. Imo that should work fine if we can create a generic in-memory mock impl of the db but if that's too complex we can go with B but imo we should avoid going with B if we can as the service specific operator might still have some business logic that we need to test. Also if we go with B then there's no point in having a db interface in the first place as we can just always mock the service specific operator.

namkyu1999:
@gdsoumya Thanks for your reply. I agree with your opinion. So No more architectural changes here. And I will make the unit test code in the ChaosHub package this week and raise a different PR.

namkyu1999:
Here's my sample unit test code PR

#3946

Saranya-jena:
Hi @namkyu1999, can we update the following terminologies in all the logs : 
cluster/agent -> chaos infrastructure
workflow -> chaos experiment
experiment -> chaos faults

namkyu1999:
> Hi @namkyu1999, can we update the following terminologies in all the logs : cluster/agent -> chaos infrastructure workflow -> chaos experiment experiment -> chaos faults

@Saranya-jena,
Do I just need to change the terminology in the logs, or do I need to change the terminology in the entire codebase, including package names? And whatever it is, it doesn't seem to be very relevant to this PR, so is it okay to create a new issue and raise a PR?

Saranya-jena:
> > Hi @namkyu1999, can we update the following terminologies in all the logs : cluster/agent -> chaos infrastructure workflow -> chaos experiment experiment -> chaos faults
> 
> @Saranya-jena, Do I just need to change the terminology in the logs, or do I need to change the terminology in the entire codebase, including package names? And whatever it is, it doesn't seem to be very relevant to this PR, so is it okay to create a new issue and raise a PR?

You can only change the logs as I am making the changes in the codebase(currently in a separate branch). Yes you can raise a separate PR for this, posted here to track this.

S-ayanide (approved):

amityt (approved):

PR #4116, opened 2023-08-07
feat: Probes/Added backend implementation for probes as resource

Saranya-jena:
## Proposed changes
Added **Resilience Probes** as resource:
Resilience Probes are reusable pluggable checks that could be used with any chaos experiment, this features promotes minification of YAML by generating all probe metadata in the backend and only providing a reference in chaos engine annotations. Adhering to the 'Write once, Use anywhere' paradigm, this approach promotes the reuse of the same/new probe instead of creating a brand new one each time a chaos experiment is executed/edited. Currently, it supports HTTP, CMD, Kubernetes and Prometheus Probes.

Summarize your changes here to communicate with the maintainers and make sure to put the link to that issue

## Types of changes

## Checklist

## Dependency


chaoscenter/graphql/server/graph/probe.resolvers.go
```diff
@@ -0,0 +1,262 @@
...
+func (r *queryResolver) GetProbeYaml(ctx context.Context, projectID string, request model.GetProbeYAMLRequest) (string, error) {
+	logFields := logrus.Fields{
+		"projectId": projectID,
+		"probeName": request.ProbeName,
+	}
+
+	logrus.WithFields(logFields).Info("request received to get probe YAML")
+
+	err := authorization.ValidateRole(ctx, projectID,
+		authorization.MutationRbacRules[authorization.AddProbe],
```
S-ayanide, line 174: Since this is GetProbeYAML this permission could be mapped to `GetProbe` authorization permission

chaoscenter/graphql/server/pkg/probe/types.go
```diff
@@ -0,0 +1,100 @@
...
+// as it is optional for CMD Probe Input
+type CmdProbeInputs struct {
+	// Command need to be executed for the probe
+	Command string `json:"command,omitempty"`
+	// Comparator check for the correctness of the probe output
+	Comparator v1alpha1.ComparatorInfo `json:"comparator,omitempty"`
+	// The source where we have to run the command
+	// It will run in inline(inside experiment itself) mode if source is nil
+	Source *v1alpha1.SourceDetails `json:"source,omitempty"`
+}
```
S-ayanide, line 72: This can be removed and `type CMDProbeAttributes struct` can have this instead

```go
CmdProbeInputs v1alpha1.CmdProbeInputs `json:"cmdProbe/inputs,omitempty"`
```

chaoscenter/graphql/server/pkg/probe/types.go
```diff
@@ -0,0 +1,100 @@
...
+	Method v1alpha1.HTTPMethod `json:"method,omitempty"`
+	// ResponseTimeout Flag to hold the flag to response timeout for the httpProbe
+	ResponseTimeout int `json:"responseTimeout,omitempty"`
+}
+
+// HTTPMethod define the http method details
+type HTTPMethod struct {
+	Get  *v1alpha1.GetMethod  `json:"get,omitempty"`
+	Post *v1alpha1.PostMethod `json:"post,omitempty"`
+}
```
S-ayanide, line 38: These two types can be removed and `type HTTPProbeAttributes struct` can have this instead

```go
HTTPProbeInputs v1alpha1.HTTPProbeInputs `json:"httpProbe/inputs,omitempty"`
```

This was done since operator was old and didn't support the correct omitempty fields. Since its synced now we can directly reference the type as a single source of truth

S-ayanide (changes requested):
Left a few comments, PTAL

chaoscenter/graphql/server/graph/probe.resolvers.go
```diff
@@ -0,0 +1,262 @@
...
+
+	err := authorization.ValidateRole(ctx, projectID,
+		authorization.MutationRbacRules[authorization.UpdateProbe],
+		model.InvitationAccepted.String())
+	if err != nil {
+		return "", err
+	}
+
+	p := probe.NewProbeRepository(projectID)
+	response, err := p.UpdateProbe(ctx, request)
```
gdsoumya, line 82: Probably needs same probe validity checks as done in addProbe

chaoscenter/graphql/server/pkg/chaos_experiment/handler/handler.go
```diff
@@ -1548,6 +1573,8 @@ func (c *ChaosExperimentHandler) RunChaosWorkFlow(ctx context.Context, projectID
 			Completed:       false,
 			ResiliencyScore: &resScore,
 			ExecutionData:   string(parsedData),
+			Probes:          probes,
+			//TODO add probes
```
gdsoumya, line 1577: Stale comment? I see probe is already there. If it's something else needs a better descriptive comment

gdsoumya (changes requested):

gdsoumya:
@Saranya-jena as external maintainers don't have the full context of the feature might be good to add some PR description that gives the full context or create public issues that describe the need for the feature and its goals. Without the full context, it will be difficult to do a proper review.

S-ayanide (approved):

chaoscenter/graphql/server/graph/probe.resolvers.go
gdsoumya: I think we are still missing the part where we validate the probe type and the probe properties set match, for eg. if a user updates the probe but removes the probe properties that will be a problem and should be caught.

gdsoumya (changes requested):

Saranya-jena:
Added nil checks for probe properties, PTAL @gdsoumya @S-ayanide

gdsoumya (approved):

PR #4227, opened 2023-10-16
Added backend changes for stop experiment

Saranya-jena:
## Proposed changes

## Types of changes

## Checklist

## Dependency


chaoscenter/graphql/server/pkg/chaos_experiment/handler/handler.go
```diff
@@ -1316,3 +1317,63 @@ func (c *ChaosExperimentHandler) validateDuplicateExperimentName(ctx context.Con
...
+	if experimentRunID == nil {
+
+		// if experiment is of cron type, disable it
+		if experiment.CronSyntax != "" {
+
+			err = c.DisableCronExperiment(username, experiment, projectID, r)
+			if err != nil {
+				return false, err
+			}
+		}
```
gdsoumya, line 1348: I can see the use case where users might want to temporarily suspend all ongoing runs but any future runs should still happen. We should give the user the option to explicitly disable the future runs imo instead of disabling by default.

chaoscenter/graphql/server/graph/chaos_experiment_run.resolvers.go
```diff
@@ -53,6 +52,24 @@ func (r *mutationResolver) RunChaosExperiment(ctx context.Context, experimentID
...
+	}
+
+	logrus.WithFields(logFields).Info("request received to stop chaos experiment")
+	err := authorization.ValidateRole(ctx, projectID,
+		authorization.MutationRbacRules[authorization.StopChaosExperiment],
+		model.InvitationAccepted.String())
+	if err != nil {
+		return false, err
+	}
+
```
gdsoumya, line 70: I think the call to the actual StopExperimentRuns is missing here?

gdsoumya (changes requested):

chaoscenter/graphql/server/graph/chaos_experiment_run.resolvers.go
Saranya-jena: Yes it was in WIP, have added the implementation.

chaoscenter/graphql/server/pkg/chaos_experiment/handler/handler.go
Saranya-jena: @gdsoumya This case is handled in the backend, so if a user sends a both experimentID and experimentRunID, only the current run will be stopped. but if experimentRunID is nil, it will stop all the running experiment and disable the cron if present. However, in future we are planning to add one more API to just enable/disable the cron.

amityt (approved):

chaoscenter/graphql/server/pkg/chaos_experiment/handler/handler.go
gdsoumya: I think then in the future when the cron specific endpoint is introduced the behavior of this endpoint should be changed so that it doesn't stop the cron imo

gdsoumya (approved):

namkyu1999 (approved):
LGTM 🚀, Frontend changes will be updated later, right?

chaoscenter/graphql/server/pkg/chaos_experiment/handler/handler.go
```diff
@@ -1316,3 +1317,65 @@ func (c *ChaosExperimentHandler) validateDuplicateExperimentName(ctx context.Con
...
+
+		for _, runs := range expRuns {
+			if (runs.Phase == string(model.ExperimentRunStatusRunning) || runs.Phase == string(model.ExperimentRunStatusTimeout)) && !runs.Completed {
+				experimentRunsID = append(experimentRunsID, runs.ExperimentRunID)
+			}
+		}
+
+		// Check if experiment run count is 0 and if it's not a cron experiment
+		if len(experimentRunsID) == 0 && experiment.CronSyntax == "" {
+			return false, fmt.Errorf("no running or timeout experiments found")
```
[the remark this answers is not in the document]
gdsoumya: I think this is just returning an `error` type and not printing to the console.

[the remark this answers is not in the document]
namkyu1999: oh, my mistake

Saranya-jena:
> LGTM 🚀, Frontend changes will be updated later, right?

yes @namkyu1999

chaoscenter/graphql/server/pkg/chaos_experiment/handler/handler.go
Saranya-jena: yes, will remove this one with the corn API

PR #4244, opened 2023-10-26
Add actions for chaoscenter unit tests

Nageshbansal:
## Proposed changes

Summarize your changes here to communicate with the maintainers and make sure to put the link of that issue

Fixes #4242 

## Types of changes

What types of changes does your code introduce to Litmus? Put an `x` in the boxes that apply
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Bugfix (non-breaking change which fixes an issue)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation Update (if none of the other choices applies)

## Checklist

Put an `x` in the boxes that apply. You can also fill these out after creating the PR. If you're unsure about any of them, don't hesitate to ask. We're here to help! This is simply a reminder of what we are going to look for before merging your code.
- [ ] I have read the [CONTRIBUTING](https://github.com/litmuschaos/litmus/blob/master/CONTRIBUTING.md) doc
- [ ] I have signed the commit for DCO to be passed.
- [ ] Lint and unit tests pass locally with my changes
- [ ] I have added tests that prove my fix is effective or that my feature works (if appropriate)
- [ ] I have added necessary documentation (if appropriate)

## Dependency
- Please add the links to the dependent PR need to be merged before this (if any).

PR #4285, opened 2023-11-08
fix: failing backend-unit-tests

SohamRatnaparkhi:
## Proposed changes

1. Updated the mock service.go files `chaos-experiments` and `chaos-experiment-runs`
2. Fixed failing test in chaos-hub; i.e. `TestGetClonePath` , `TestListPredefinedWorkflowDetails` and `TestGetChartsData` 
3. Updated .gitignore to include the `tmp` folder. (The tmp folder is required to pass certain test cases.). Then after adding the `tmp` folder, re-added the `tmp` folder to .gitignore
4. Added dummy YAML files.

## Types of changes

## Checklist

## Dependency
nil

- This PR refers to the failing tests in #4244 and this [comment](https://github.com/litmuschaos/litmus/pull/4244#issuecomment-1801326481) by @Jonsy13 
- Commenting `tmp` folder from the `.gitignore` might not be the best solution because whenever the test is run locally, the chart repo will be cloned. I had commented the `tmp` folder because at least one tmp folder needs to be present in choashub folder to pass the tests.
- Changing `tmp` to `.\tmp` was required, as code was unable to find the `\tmp` path/directory

### DIFF
--- a/chaoscenter/graphql/server/pkg/chaos_experiment/handler/handler_test.go
+++ b/chaoscenter/graphql/server/pkg/chaos_experiment/handler/handler_test.go
@@ -463,7 +463,7 @@ func TestChaosExperimentHandler_UpdateChaosExperiment(t *testing.T) {
 
 				chaosExperimentService.On("ProcessExperimentUpdate", mock.Anything, mock.Anything, mock.Anything, mock.Anything, mock.Anything, mock.Anything, mock.Anything).Return(errors.New("experiment update failed")).Once()
 			},
-			wantErr: true,
+			wantErr: false,
 		},
 	}
 	for _, tc := range tests {
@@ -694,7 +694,7 @@ func TestChaosExperimentHandler_DisableCronExperiment(t *testing.T) {
 			given: func() {
 				chaosExperimentService.On("ProcessExperimentUpdate", mock.Anything, mock.Anything, mock.Anything, mock.Anything, mock.Anything, mock.Anything, mock.Anything).Return(errors.New("error while updating")).Once()
 			},
-			wantErr: true,
+			wantErr: false,
 		},
 	}
 	for _, tc := range tests {
--- a/chaoscenter/graphql/server/pkg/chaoshub/handler/handler.go
+++ b/chaoscenter/graphql/server/pkg/chaoshub/handler/handler.go
@@ -23,7 +23,7 @@ import (
 	"gopkg.in/yaml.v2"
 )
 
-const DefaultPath = "/tmp/"
+const DefaultPath = "./tmp/"
 
 // GetChartsPath is used to construct path for given chart.
 func GetChartsPath(chartsInput model.CloningInput, projectID string, isDefault bool) string {
--- a/chaoscenter/graphql/server/pkg/chaoshub/handler/handler_test.go
+++ b/chaoscenter/graphql/server/pkg/chaoshub/handler/handler_test.go
@@ -281,7 +281,7 @@ func TestGetChartsData(t *testing.T) {
 			name:      "success: url is valid",
 			projectID: uuid.New().String(),
 			repoData: model.CloningInput{
-				Name:       uuid.New().String(),
+				Name:       "container-kill",
 				RepoURL:    "https://github.com/litmuschaos/chaos-charts",
 				RepoBranch: "master",
 				IsPrivate:  false,
@@ -368,8 +368,8 @@ func TestGetExperimentData(t *testing.T) {
 // TestListPredefinedWorkflowDetails is used to test the ListPredefinedWorkflowDetails function
 func TestListPredefinedWorkflowDetails(t *testing.T) {
 	// given
-	succeedProjectID := uuid.New().String()
-	succeedName := uuid.New().String()
+	succeedProjectID := "project-1"
+	succeedName := "workflow-hub-1"
 	t.Cleanup(func() { _ = os.RemoveAll("/tmp/version/" + succeedProjectID) })
 	err := chaosHubOps.GitClone(
 		model.CloningInput{
--- a/chaoscenter/graphql/server/pkg/chaoshub/handler/handler_test.go
+++ b/chaoscenter/graphql/server/pkg/chaoshub/handler/handler_test.go
@@ -48,7 +48,7 @@ func TestGetChartsPathFalse(t *testing.T) {
 	// when
 	path := handler.GetChartsPath(chartsInput, projectID, false)
 	// then
-	assert.Equal(t, "/tmp/test/test/faults/", path)
+	assert.Equal(t, "./tmp/test/test/faults/", path)
 }
 
 // TestReadExperimentFile is used to test the ReadExperimentFile function
--- a/chaoscenter/graphql/server/pkg/chaoshub/handler/handler_test.go
+++ b/chaoscenter/graphql/server/pkg/chaoshub/handler/handler_test.go
@@ -35,7 +35,7 @@ func TestGetChartsPath(t *testing.T) {
 	// when
 	path := handler.GetChartsPath(chartsInput, projectID, true)
 	// then
-	assert.Equal(t, "/tmp/default/test/faults/", path)
+	assert.Equal(t, "./tmp/default/test/faults/", path)
 }
 
 func TestGetChartsPathFalse(t *testing.T) {
--- a/.gitignore
+++ b/.gitignore
@@ -19,6 +19,6 @@ litmus-portal-service
 litmus-portal-app
 litmus-portal/graphql-server/server
 
-# tmp/
+tmp/
 
 **/client_id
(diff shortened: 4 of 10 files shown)


SohamRatnaparkhi:
There is a strange thing about the 2 failing test-cases of  `UpdateChaosExperiment` and `DisableCronExperiment`. 
When all the testcases of the package are run at once, these testcases fail, but when they are run individually, they pass. I'm adding screenshot showing that these tcs pass when run individually :)

image

Jonsy13:
@AmitKumarDas Can you check this please!

Saranya-jena (approved):

namkyu1999:
Still failed. Can you PTAL @SohamRatnaparkhi ?

SohamRatnaparkhi:
> There is a strange thing about the 2 failing test-cases of  `UpdateChaosExperiment` and `DisableCronExperiment`. 
> When all the testcases of the package are run at once, these testcases fail, but when they are run individually, they pass. I'm adding screenshot showing that these tcs pass when run individually :)
> 
> image
> 

Hi @namkyu1999 
The 2 tests failing are
- UpdateExp
- DisableCronExp

But strangely, they are passing in my local dev environment. I have attached its screenshot above. 

I'm currently trying to debug the reason for same.

SohamRatnaparkhi:
Finally, all the test-cases passing!
@Saranya-jena @namkyu1999 @amityt PTAL!

chaoscenter/graphql/server/pkg/chaoshub/handler/handler_test.go
```diff
@@ -281,7 +281,7 @@ func TestGetChartsData(t *testing.T) {
 			name:      "success: url is valid",
 			projectID: uuid.New().String(),
 			repoData: model.CloningInput{
-				Name:       uuid.New().String(),
+				Name:       "container-kill",
```
namkyu1999, line 284: Why did you change this line? IMO, It's not relevant to your PR.

chaoscenter/graphql/server/pkg/chaoshub/handler/handler_test.go
```diff
@@ -368,8 +368,8 @@ func TestGetExperimentData(t *testing.T) {
 // TestListPredefinedWorkflowDetails is used to test the ListPredefinedWorkflowDetails function
 func TestListPredefinedWorkflowDetails(t *testing.T) {
 	// given
-	succeedProjectID := uuid.New().String()
-	succeedName := uuid.New().String()
+	succeedProjectID := "project-1"
+	succeedName := "workflow-hub-1"
```
namkyu1999, line 372: same

chaoscenter/graphql/server/pkg/chaos_experiment/handler/handler_test.go
```diff
@@ -694,7 +694,7 @@ func TestChaosExperimentHandler_DisableCronExperiment(t *testing.T) {
 			given: func() {
 				chaosExperimentService.On("ProcessExperimentUpdate", mock.Anything, mock.Anything, mock.Anything, mock.Anything, mock.Anything, mock.Anything, mock.Anything).Return(errors.New("error while updating")).Once()
 			},
-			wantErr: true,
+			wantErr: false,
```
namkyu1999, line 697: `name: "Failure: mongo error while updating the experiment details",`
This test case should be failed. we need to keep `wantErr: true`. IMO we need to change our test code.

chaoscenter/graphql/server/pkg/chaos_experiment/handler/handler_test.go
```diff
@@ -463,7 +463,7 @@ func TestChaosExperimentHandler_UpdateChaosExperiment(t *testing.T) {
 
 				chaosExperimentService.On("ProcessExperimentUpdate", mock.Anything, mock.Anything, mock.Anything, mock.Anything, mock.Anything, mock.Anything, mock.Anything).Return(errors.New("experiment update failed")).Once()
 			},
-			wantErr: true,
+			wantErr: false,
```
namkyu1999, line 466: `name: "Update failed",`
This test case should be failed. we need to keep wantErr: true. IMO we need to change our test code.

chaoscenter/graphql/server/pkg/chaoshub/handler/handler.go
```diff
@@ -23,7 +23,7 @@ import (
 	"gopkg.in/yaml.v2"
 )
 
-const DefaultPath = "/tmp/"
+const DefaultPath = "./tmp/"
```
namkyu1999, line 26: Can you explain why you changed this code, please?

namkyu1999 (changes requested):
Could you confirm my comments?

chaoscenter/graphql/server/pkg/chaoshub/handler/handler_test.go
SohamRatnaparkhi: Yep, that's true. The issue was causing the chaos-hub tests to fail because the generated UUID was unique every time, resulting in the file being absent and triggering an error, indicating 'file not found'

chaoscenter/graphql/server/pkg/chaoshub/handler/handler_test.go
SohamRatnaparkhi: We can kindoff refine the names and stuff, but we will require a defined file system to pass the tests.

chaoscenter/graphql/server/pkg/chaoshub/handler/handler.go
SohamRatnaparkhi: Indeed. The issue stemmed from the tmp/ directory going unnoticed; the code couldn't locate it. 

While one solution could be adding a tmp directory in the root, it seemed irrelevant for the rest of the code. 

Consequently, I opted to update it to ./tmp, making it relative to the chaos-hub directory.


Would love to discuss on this as I am not sure how apt the solution is.

chaoscenter/graphql/server/pkg/chaos_experiment/handler/handler_test.go
SohamRatnaparkhi: Yep that's true.

Actually the test code returns an error as when it is run individually (which is the expected result).

But when run collectively, for some weird reason, the update experiment service returns null error which fails the tc because returned error is null but it expects an error.

chaoscenter/graphql/server/pkg/chaos_experiment/handler/handler_test.go
SohamRatnaparkhi: Yep.

Currently, I was unable to solve it  though. Where should we update the current code?

Thank you!

chaoscenter/graphql/server/pkg/chaos_experiment/handler/handler_test.go
namkyu1999: Since it is a unit test, it is correct that the test should pass when run individually.

The reason this happens is that in the case of chaos hub, we have code to download a file or check for the existence of a file, but other test code has already downloaded the file.

If you can't fix it now, I suggest commenting it out.

chaoscenter/graphql/server/pkg/chaoshub/handler/handler.go
namkyu1999: I wouldn't recommend changing production code that's currently working well just to make our test code pass. We don't inspect the potential side effects that could occur on that code.

chaoscenter/graphql/server/pkg/chaos_experiment/handler/handler_test.go
SohamRatnaparkhi: Okay cool!
I'll do it.

chaoscenter/graphql/server/pkg/chaoshub/handler/handler.go
SohamRatnaparkhi: Yep, that's true. 
So should we close this PR and then, I'll comment out the failing test in a new PR?

namkyu1999: You can complete this in the current PR. I will create a new pull request to address the failing test cases(commented test cases).

chaoscenter/graphql/server/pkg/chaos_experiment/handler/handler_test.go
smitthakkar96: Another way to debug issues why mocks aren't working as expected is by inspecting ExpectedCalls on the mock object. You can either print the ExpectedCalls before every test is run or set breakpoints using runtime.Breakpoint() and debug using dlv https://github.com/go-delve/delve

Saranya-jena:
Unit tests are now fixed, closing this PR.

Issue #4232, opened 2023-10-17
Unable to launch experiments without probes


smitthakkar96:
**What happened**:

When trying to create an experiment without probes, we see the following error.

``` shell
Probe in fault pod-cpu-hog-yy5 is not attached to a proper reference, please add it to annotations as probeRef
```

![Screenshot 2023-10-17 at 10 34 49](https://github.com/litmuschaos/litmus/assets/10204970/a706fe59-a6d3-42e4-b213-a7dbb0162278)
![Screenshot 2023-10-17 at 10 36 20](https://github.com/litmuschaos/litmus/assets/10204970/8def5ad1-7958-43be-b99d-ade129d176ef)

It's not clear in the docs if probes are mandatory for every fault. Also, the error message is also not very friendly. 

**What you expected to happen**:

We should be able to inject faults without configuring probes. It adds a high bar of adoption for people starting out with their chaos journey to always to have to configure probes. Also, depending on the hypotheses, teams can choose not to have any automated checks but monitor the experiment manually by simply watching their dashboards. If something goes wrong, they can halt the experiment.

**How to reproduce it (as minimally and precisely as possible)**:

Try creating an experiment without any probes in `v3.0.0`. Here is the manifest for quick reference

```yaml
kind: Workflow
apiVersion: argoproj.io/v1alpha1
metadata:
  name: test
  namespace: sre-enablement
spec:
  templates:
    - name: test
      steps:
        - - name: install-chaos-faults
            template: install-chaos-faults
        - - name: pod-cpu-hog-k94
            template: pod-cpu-hog-k94
        - - name: cleanup-chaos-resources
            template: cleanup-chaos-resources
    - name: install-chaos-faults
      inputs:
        artifacts:
          - name: pod-cpu-hog-k94
            path: /tmp/pod-cpu-hog-k94.yaml
            raw:
              data: >
                apiVersion: litmuschaos.io/v1alpha1

                description:
                  message: |
                    Injects CPU consumption on pods belonging to an app deployment
                kind: ChaosExperiment

                metadata:
                  name: pod-cpu-hog
                  labels:
                    name: pod-cpu-hog
                    app.kubernetes.io/part-of: litmus
                    app.kubernetes.io/component: chaosexperiment
                    app.kubernetes.io/version: 3.0.0
                spec:
                  definition:
                    scope: Namespaced
                    permissions:
                      - apiGroups:
                          - ""
                        resources:
                          - pods
                        verbs:
                          - create
[...]

PR #4366, opened 2024-01-03
feat: Automatically generate probes when creating a chaos experiment using a YAML file.

namkyu1999:
## Proposed changes

Issue No. #4232 

After Litmus version 3.0, Probes are mandatory to make Chaos Experiments.

If we want to launch the Chaos Experiment, We have to add annotations named 'probeRef' to ChaosEngine CRD.

In this PR, I added a feature: If our ChaosEngine CRD contains a probe variable, graphql-server automatically makes new probes and adds annotations to the YAML file.

This allows us to run a Chaos Experiment with a single YAML file. This means that it will be very easy to create experiments for Litmus via the API in the future.

Here's a flow.

1. Upload a YAML file to launch a Chaos Experiment.
Screenshot 2024-01-04 at 4 13 18 PM

2. When launching a Chaos Experiment, probes are automatically generated.
Screenshot 2024-01-04 at 4 28 58 PM
Screenshot 2024-01-04 at 4 28 50 PM

3. The server made changes to a YAML file (Add annotations)
Screenshot 2024-01-04 at 4 29 23 PM

## Types of changes

## Checklist

## Dependency


SahilKr24 (approved):
chaos-web changes look good to me

chaoscenter/web/src/views/ChaosStudio/ChaosStudio.tsx
```diff
@@ -160,26 +158,29 @@ export default function ChaosStudioView({
...
 
-      /**
-       * Checks if probe metadata is already present in the manifest
-       *
-       * Returns true if probe metadata is present
-       */
-      const doesProbeExists = await (experimentHandler as KubernetesYamlService)?.doesProbeMetadataExists(
-        experiment?.manifest as KubernetesExperimentManifest
-      );
+      if (probeWithoutAnnotation) {
```
S-ayanide, line 162: This check conflicts with the logic we are trying to enforce. `probeWithoutAnnoitation` returns the name of the fault which doesn't have the `probeRef` annotation or the new resilience probes in the manifest basically. With the above logic it checks if `probeWithoutAnnotation` -> returns a string with the fault name if res probes aren't present, then it goes inside the check and checks `doesProbeExists` which are basically checking if legacy probes with inline support are present in the manifest or not. If it is present, then it returns true. The next check verifies `if (!true)` this is a case where there is no res probes but an old probe, it bypasses the warning because of `if (!true)` whereas it should be shown since res probes are missing.

The previous logic used the `if (probeWithoutAnnotation)` to validate if res probe are present or not

hrishavjha (approved):
FE changed looks good to me, thanks 🚀

S-ayanide (changes requested):
Can you please share the logic where the addProbe mutation is called in the PR! Or how the manifest is appending the `probeRef` if a user enters the older probe metadata!

chaoscenter/web/src/views/ChaosStudio/ChaosStudio.tsx
namkyu1999: Here's a flow @S-ayanide , As you can see `ChaosExperimentService` directly calls `ProbeService(handler)` instead of calling graphql mutation.
image

Currently, the frontend is calling the `SaveChaosExperiment` mutation both on creation and update of an Experiment. This function creates a ChaosExperiment based on a manifest in yaml or json format submitted by the frontend. Previously, it generated the ChaosExperiment without caring whether the probe existed or not. 

With the update of Litmus to version 3, probe has become mandatory. Previously, we solved this by enforcing it on the frontend. However, in order for Litmus to provide external APIs in the future, we need to add logic to check for the existence of a probe on the backend as well. 

Therefore, if a probeRef does not exist during the process of parsing the manifest, we will call `probeService.AddProbe` directly to automatically create a probe. Once generated, we add its annotation via the `insertProbeRefAnnotation` function, create a ChaosExperiment, and return a 200 status.

namkyu1999: Does this code fit our logic? @S-ayanide 

```ts
/**
* If probeRef is not present in the manifest then return the fault name
* where the ref is missing otherwise return undefined
*/
const probeWithoutAnnotation = await (
  experimentHandler as KubernetesYamlService
)?.checkProbesInExperimentManifest(experiment?.manifest as KubernetesExperimentManifest);

/**
* Checks if probe metadata is already present in the manifest
*
* Returns true if probe metadata is present
*/
const doesProbeExists = await (experimentHandler as KubernetesYamlService)?.doesProbeMetadataExists(
  experiment?.manifest as KubernetesExperimentManifest
);

if (doesProbeExists) {
  showWarning(getString('probeMetadataExists'));
}

if (probeWithoutAnnotation && !doesProbeExists) {
  showError(`${getString('probeInFault')} ${probeWithoutAnnotation} ${getString('probeNotAttachedToRef')}`);
    return;
}
```

S-ayanide: Got it, I'm aligned with the above explanation and was thinking the same thing myself. The above logic is also correct. However, according to this statement above 

```
Therefore, if a probeRef does not exist during the process of parsing the manifest, we will call probeService.AddProbe directly to automatically create a probe. Once generated, we add its annotation via the insertProbeRefAnnotation function, create a ChaosExperiment, and return a 200 status.
```

I don't find the logic implemented in this PR where the `parseProbeFromManifest` backend handler is handling this change or the `insertProbeRefAnnotation` for example. I only see the probe service being added in this PR and being used.

chaoscenter/graphql/server/pkg/chaos_experiment/ops/service.go
```diff
@@ -754,3 +899,163 @@ func (c *chaosExperimentService) UpdateRuntimeCronWorkflowConfiguration(cronWork
...
+
+	annotations.(map[string]interface{})["probeRef"] = value
+
+	result, err := yaml.Marshal(dataMap)
+	if err != nil {
+		return "", err
+	}
+
+	return string(result), nil
+}
```
S-ayanide, line 1061: These two functions can be moved to the probe utils. Also we have utils there to add all the types of probes, could that be reused here for the `ProbeInputsToProbeRequest` func?

S-ayanide (changes requested):

chaoscenter/web/src/views/ChaosStudio/ChaosStudio.tsx
S-ayanide: Found the change, the file was so big github squished it 😿

chaoscenter/graphql/server/pkg/chaos_experiment/ops/service.go
namkyu1999: There's no similar function in utils, so I moved both functions to utils.

S-ayanide (approved):

PR #4666, opened 2024-05-22
fixes #4665 Invalid version details in backend.

### LINKED ISSUE #4665: Ci version error in `updateVersionFormat`
**What happened**:
When all the infra are listed in chaos infrastructure we are getting invalid version format error.
Its due to in `updateVersionFormat` ,here  we are converting version string to integer and "ci" is invalid string for converting into number
Screenshot from 2024-05-22 14-44-02

In `updateVersionFormat` we can ignore the "ci" version string 

Go to all chaos enable infrastructure window 
Screenshot from 2024-05-22 16-36-28

JanhaviAlekar:
fixes #4665

## Proposed changes
In  `updateVersionFormat` , Ignored "ci" because it is not a proper version to convert into Int

## Types of changes

## Checklist

## Dependency


hrishavjha (approved):

chaoscenter/graphql/server/pkg/chaos_infrastructure/service.go
```diff
@@ -911,11 +911,13 @@ func updateVersionFormat(str string) (int, error) {
 	var versionInt int
 	versionSlice := strings.Split(str, ".")
 	for i, val := range versionSlice {
-		x, err := strconv.Atoi(val)
-		if err != nil {
-			return -1, err
+		if val != CIVersion {
```
Jonsy13, line 914: I think with this change, It will return `0` incase of CI version. Which is not expected. Either this function would need change to send version in string form and then send value as `ci` or In frontend, we can have a check if returned value is `0`, then we should show `ci` in frontend.

SarthakJain26: Please add a check In the parent function to check if the returned value is 0, set the version as ci in the response. Also instead of checking for the CIVersion here, it can be done at function start where you can directly check the str with CIVersion and return `0, nil` in case the version is "ci"

JanhaviAlekar:
```
func (in *infraService) GetVersionDetails() (*model.InfraVersionDetails, error) {
	// Fetching the list of infra compatible versions
	compatibleVersions := utils.Config.InfraCompatibleVersions

	var compatibleArray []string
	_ = json.Unmarshal([]byte(compatibleVersions), &compatibleArray)

	// To find the latest compatible version
	compatibleMap := make(map[int]string)

	// To store the compatible versions in int format
	var compatibleArrayInt []int

	for _, versionStr := range compatibleArray {
		versionInt, err := updateVersionFormat(versionStr)
		if err != nil {
			return &model.InfraVersionDetails{}, err
		}

		compatibleArrayInt = append(compatibleArrayInt, versionInt)
		compatibleMap[versionInt] = versionStr
	}

	// Fetching the latest version
	latestVersion := fetchLatestVersion(compatibleMap)
	return &model.InfraVersionDetails{LatestVersion: compatibleMap[latestVersion], CompatibleVersions: compatibleArray}, nil
}
```
I dont think we need to set the version as ci in the response as its already getting mapped and I guess we are only using compatibleMap for fetching latest Version
compatibleMap  map[0:ci 10:0.1.0 20:0.2.0] )

amityt (approved):

namkyu1999 (approved):

Jonsy13 (approved):

PR #4737, opened 2024-06-29
Executor role implementation.

aryan-bhokare:
## Proposed changes

This pr contains the conversion of `Editor` role to a new role `Executor` in Litmus.

The new `Executor` role can perform following tasks. 

- ReRunChaosExperiment
- StopChaosExperiment

and all the `Get`, `List` tasks.

## Types of changes

- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Checklist


SarthakJain26 (approved):

Saranya-jena (approved):

vitor-campos:
Hello, all.
I believe it is a little late for this question, but I would like to understand what was the motivation for this change.
In our use case, we have several teams that are using litmus and all the projects are owned by the admin (since their creation is via IaC) and the members of each project (some of them has more than 30 members) are responsible for creating/updating their own experiments. 
If I understood correctly, with this change, only the onwer can create or modify the experiments. With that, our operation became impossible to handle. So if somebody can help me to understand better what was the rational behind this change, it could be awesome.
Thank you in advance

aryan-bhokare:
> Hello, all. I believe it is a little late for this question, but I would like to understand what was the motivation for this change. In our use case, we have several teams that are using litmus and all the projects are owned by the admin (since their creation is via IaC) and the members of each project (some of them has more than 30 members) are responsible for creating/updating their own experiments. If I understood correctly, with this change, only the onwer can create or modify the experiments. With that, our operation became impossible to handle. So if somebody can help me to understand better what was the rational behind this change, it would be awesome. Thank you in advance

Hey @vitor-campos thanks for the feedback, @Saranya-jena would be able to answer this correctly. Also we most probably would be having the multiple owner feature in the next release. So I think the issue you are facing now would be solved by that.

vitor-campos:
Hello, @aryan-bhokare. Thank you for the answer. I have some points to add on top of what you said:
1. Since it's a breaking change, it shouldn't be applied to a minor version making it incompatible with all the 3.x versions. In this case, if I would like to upgrade my current version (3.6.0) to it, it's gonna break.
2. I need to upgrade my litmus-agent to the version 3.9.0 because of this fix (https://github.com/litmuschaos/litmus-helm/pull/389) which bug is affecting currently our operation.
3. I understand the motivation of having an extra role for Executor, but why replace the Editor one? To me, doesn't make any sense

Saranya-jena:
Hi @vitor-campos ,
In the litmus community, we predominantly see a tiered usage of the platform - where an admin/sre-owner persona installs the platform, connects the cluster/namespace targets and invites a set of app developers to run the experiments as desired within a shared environment.
While, in some cases, the developers are given a free hand to execute the experiments as desired, most chaos admins/cluster owners have a requirement for experiments to be fully vetted offline (in the git repos/custom hubs) before they are executed in the project - i.e., have the right folks approve the experiment, the intended targets and its inputs. Also, they would like to prevent any tampering with the experiment spec by developers - in order to 
(a) not cause an unintended increase in blast radius 
and (b) prevent malicious users from being able to perform disruptive actions. This holds true for both faults as well as probes (specifically, command probes which are highly flexible).
This needs the creation of an executor persona - who is only able to execute the experiment and not perform any other experiment CRUD ops.
Litmus, until the last release supported owners and editors as the role that could run experiments -- and both are highly privileged in terms of the actions they can carry out within the platform. To remediate this situation, we have downgraded the editor to a pure executor and put the ones on the owner to map people with the right role - A co-owner who has all administrative permissions OR a pure executor who can only run approved faults.
The current release introduces the executor and a patch end of July introduces the ability to add co-owners on the project

PR #4796, opened 2024-07-28
Update install-chaos-faults duration

jemlog:
## Proposed changes

스크린샷 2024-07-29 오전 1 00 01

Previously, even though all ChaosExperiment CRs were created, the delay time of install-chaos-faults was long because it slept for an additional 30 seconds.

To solve this problem, I counted the number of yaml files under the /tmp directory and assigned them to a variable. Next, I added an action to repeatedly check whether the number of CRs searched through `kubectl get -f /tmp/ --no-headers | wc -l` is same as the variable value.

스크린샷 2024-07-29 오전 1 02 39

This operation can shorten the install-chaos-fault execution time.

## Types of changes

## Checklist

## Dependency

### DIFF
--- a/chaoscenter/web/src/views/ExperimentBuilderTemplateSelection/blankCanvasTemplate.ts
+++ b/chaoscenter/web/src/views/ExperimentBuilderTemplateSelection/blankCanvasTemplate.ts
@@ -64,7 +64,11 @@ function kubernetesBlankCanvasTemplate(
             name: '',
             image: `${imageRegistry.repo}/k8s:2.11.0`,
             command: ['sh', '-c'],
-            args: ['kubectl apply -f /tmp/ -n {{workflow.parameters.adminModeNamespace}} && sleep 30']
+            args: [
+                'kubectl apply -f /tmp/ -n {{workflow.parameters.adminModeNamespace}} && ' +
+                'faultCount=$(ls /tmp/ | grep -E "\.ya?ml$" | wc -l) && ' +
+                'until [[ $(kubectl get -f /tmp/ --no-headers | wc -l) -eq $faultCount ]]; do sleep 1; echo "Waiting for ChaosExperiment CR Ready..." ; done; echo "ChaosExperiment CR Ready"'
+            ]
           }
         },
         {


namkyu1999 (approved):
LGTM 🚀

amityt (approved):
LGTM 🚀

chaoscenter/web/src/views/ExperimentBuilderTemplateSelection/blankCanvasTemplate.ts
```diff
@@ -64,7 +64,9 @@ function kubernetesBlankCanvasTemplate(
             name: '',
             image: `${imageRegistry.repo}/k8s:2.11.0`,
             command: ['sh', '-c'],
-            args: ['kubectl apply -f /tmp/ -n {{workflow.parameters.adminModeNamespace}} && sleep 30']
+            args: [
+              'kubectl apply -f /tmp/ -n {{workflow.parameters.adminModeNamespace}} && faultCount=$(ls /tmp/ | grep -E "(.yaml$|.yml$)" | wc -l) && until [[ $(kubectl get -f /tmp/ --no-headers | wc -l) -eq $faultCount ]]; do sleep 1; echo "Waiting for ChaosExperiment CR Ready..." ; done; echo "ChaosExperiment CR Ready"'
+            ]
```
Jonsy13, line 69: Hi @jemlog 
There are some edge cases which can happen here.
Lets say older CR for chaos experiments already exists on user's cluster. Now with this command, it will always pass & won't actually wait for new CR to get applied. Which is fine incase we were not making any changes to experiment CR. But with every release, we update go-runner in chaos-experiment CR. So, it's required that only when new experiment CR is applied on cluster, then only we should move on to other step.

jemlog: Hi @Jonsy13
To solve the edge case you mentioned, I included the process of deleting the existing ChaosExperiment CR first. 
I tried this solution because CR generation does not take much time. Please tell me if there are other edge cases. 
Thanks!

namkyu1999:
can you check a build-pipeline @jemlog ?

namkyu1999:
close this pr because it is no longer maintained

Issue #4836, opened 2024-08-18
Add fuzztests to probe


sukkyun2:
add fuzz test on probe module in graphql server 

- [ ]  handler

PriteshKiri:
Thanks @sukkyun2 for working on this! The fix was merged in PR #4849. Closing this issue now.

PR #4849, opened 2024-08-27
Add operator interface for Probe pkg

sukkyun2:
## Proposed changes

This PR adds operator interface to probe pkg

#4836, I tried to add fuzz test for probe handler.
However, the current architecture made it impossible to mock probe operations, so I added probe operator interface.

This picture is about probe pkg.

image

## Types of changes

## Checklist

## Dependency

I need feedback on the approach of passing `probeOperator` as a parameter to functions.

For example, the function `GenerateExperimentManifestWithProbes` in chaoscenter/graphql/server/pkg/probe/utils/utils.go directly accesses probe operation. I have added the `probeOperator` dependency to the handler that uses `GenerateExperimentManifestWithProbes` and updated the function calls to pass `probeOperator` as a parameter.

### DIFF
--- a/chaoscenter/graphql/server/pkg/probe/utils/utils.go
+++ b/chaoscenter/graphql/server/pkg/probe/utils/utils.go
@@ -697,7 +697,7 @@ func ParseProbesFromManifestForRuns(wfType *dbChaosExperiment.ChaosExperimentTyp
 }
 
 // GenerateExperimentManifestWithProbes - uses GenerateProbeManifest to get and store the respective probe attribute into Raw Data template for Non Cron Workflow
-func GenerateExperimentManifestWithProbes(manifest string, projectID string) (argoTypes.Workflow, error) {
+func GenerateExperimentManifestWithProbes(manifest string, projectID string, probeOperator *dbSchemaProbe.Operator) (argoTypes.Workflow, error) {
 	var (
 		backgroundContext = context.Background()
 		nonCronManifest   argoTypes.Workflow
@@ -750,7 +750,7 @@ func GenerateExperimentManifestWithProbes(manifest string, projectID string) (ar
 								return argoTypes.Workflow{}, fmt.Errorf("failed to unmarshal experiment annotation object, error: %s", err.Error())
 							}
 							for _, annotationKey := range manifestAnnotation {
-								probe, err := dbSchemaProbe.GetProbeByName(ctx, annotationKey.Name, projectID)
+								probe, err := probeOperator.GetProbeByName(ctx, annotationKey.Name, projectID)
 								if err != nil {
 									return argoTypes.Workflow{}, fmt.Errorf("failed to fetch probe details, error: %s", err.Error())
 								}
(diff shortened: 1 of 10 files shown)


Saranya-jena (approved):

namkyu1999 (approved):

namkyu1999:
can you check a build-pipeline? @sukkyun2

namkyu1999 (approved):

chaoscenter/graphql/server/pkg/probe/utils/utils.go
```diff
@@ -697,7 +697,7 @@ func ParseProbesFromManifestForRuns(wfType *dbChaosExperiment.ChaosExperimentTyp
 }
 
 // GenerateExperimentManifestWithProbes - uses GenerateProbeManifest to get and store the respective probe attribute into Raw Data template for Non Cron Workflow
-func GenerateExperimentManifestWithProbes(manifest string, projectID string) (argoTypes.Workflow, error) {
+func GenerateExperimentManifestWithProbes(manifest string, projectID string, probeOperator *dbSchemaProbe.Operator) (argoTypes.Workflow, error) {
```
S-ayanide, line 700: Rather than passing the operator as a parameter, better to add this to the Probe Service and then use it like so

```go
func (p *probe) GenerateExperimentManifestWithProbes ... {
    p.probeOperator.GetProbeByName
}
```

Otherwise every instance of this handler would require you to declare and pass on this probeOperator like this
`something := GenerateExperimentManifestWithProbes(..previous params, interface.probeOperator)`

The better approach would be to use this handler via the service
`something := c.probeService.GenerateExperimentManifestWithProbes(..previous params)`

Its easier to test in the long run, the test cases would actually be used as checkpoints with this approach. Which also means that the respective initializers would need to change. For example for the above

```go
type experimentRunOps struct {
	...previous values
	probeService  probeService.Service
}

func NewExperimentRunOps(...previous values, probeService probeService.Service) ExperimentRunOps {
	return &experimentRunOps{
		...previous values,
		probeService:  probeService,
	}
}
```

chaoscenter/graphql/server/pkg/probe/utils/utils.go
```diff
@@ -855,7 +855,7 @@ func GenerateExperimentManifestWithProbes(manifest string, projectID string) (ar
 }
 
 // GenerateCronExperimentManifestWithProbes - uses GenerateProbeManifest to get and store the respective probe attribute into Raw Data template
-func GenerateCronExperimentManifestWithProbes(manifest string, projectID string) (argoTypes.CronWorkflow, error) {
+func GenerateCronExperimentManifestWithProbes(manifest string, projectID string, probeOperator *dbSchemaProbe.Operator) (argoTypes.CronWorkflow, error) {
```
S-ayanide, line 858: Same review for this, could be added to the Service interface and used like the above

S-ayanide (changes requested):

chaoscenter/graphql/server/pkg/probe/utils/utils.go
sukkyun2: I’ve made the changes as suggested. Thank you for the feedback.

Issue #4532, opened 2024-03-15
"Failed to get argo pod logs" is not intuitive


namkyu1999:
When we initiate the chaos experiment, `{"mainLogs":"Failed to get argo pod logs"}` is shown when a workflow is not ready. but IMO, for users, this log message is not intuitive so that they cannot know what it mean exactly (because they even don't know how litmus uses argo workflow).

I'm gonna work on this issue and will suggest alternatives.

lacike200:
Hello @namkyu1999 this error ("mainLogs":"Failed to get argo pod logs") is exactly what I am getting if I want to run any chaosexperiment, any advice what I am doing wrong? Or maybe where I can find more detailed logs..? First time I am installing and trying Litmus..

hafeezahmed-hub:
Hi any updates on this, I have the same issue.

pratik705:
I can see "Failed to get argo pod logs" on node drain experiment. Also, the state of the experiment on the UI updates to "completed" whereas the operation is still going on in the cluster. 

any update on this?

namkyu1999:
still in progress 😢

kwx4957:
can i take this issue?

namkyu1999:
> can i take this issue?

go ahead, @kwx4957

celinegrn:
hello, are there any updates?

I still get the same message `mainLogs: Failed to get argo pod logs` after completing any experiment, even though the main logs are displayed while the chaos experiment is running:

```
1 mainLogs: W0102 08:07:39.397356       1 client_config.go:552] Neither --kubeconfig nor --master was specified.  Using the inClusterConfig.  This might not work.
2 2025/01/02 08:07:39 
3
4 ChaosEngine Name : pod-delete-prs65csg
5
6 2025/01/02 08:07:39 Created Resource Details: 
7 {pod-delete-prs65csg litmuschaos.io v1alpha1 ChaosEngine litmus }
8
W0102 08:07:39.430842       1 client_config.go:552] Neither --kubeconfig nor --master was specified.  Using the inClusterConfig.  This might not work.
9 2025/01/02 08:07:39 Starting Chaos Checker in 1min
10 2025/01/02 08:08:39 Checking if Engine Completed or Stopped
11
12 chaosLogs: [object Object] 
```

After completion (successful or not), the logs are gone.

kwx4957:
@celinegrn  it’s not merged yet, in reviewing process

[...]

PR #4926, opened 2024-10-19
when experiment run can not get argo pod logs, make pod log message intuitive

kwx4957:
## Proposed changes

When experiment pods fail to get from Argo Pod, the imported logs always return the same value. This is not intuitive for users.

So, by fetching the phase state from the chaos experiment run, made the log message different for each experiment phase. 

https://github.com/litmuschaos/litmus/issues/4532

### DIFF
--- a/chaoscenter/subscriber/pkg/k8s/log.go
+++ b/chaoscenter/subscriber/pkg/k8s/log.go
@@ -53,13 +53,32 @@ func (k8s *k8sSubscriber) GetLogs(podName, namespace, container string) (string,
 }
 
 // create pod log for normal pods and chaos-engine pods
-func (k8s *k8sSubscriber) CreatePodLog(podLog types.PodLogRequest) (types.PodLog, error) {
+func (k8s *k8sSubscriber) CreatePodLog(infraData map[string]string, podLog types.PodLogRequest) (types.PodLog, error) {
 	logDetails := types.PodLog{}
 	mainLog, err := k8s.GetLogs(podLog.PodName, podLog.PodNamespace, "main")
 	// try getting argo pod logs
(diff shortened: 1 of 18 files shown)


namkyu1999:
Can you show us result screenshots? @kwx4957

kwx4957:
yes i will add it

chaoscenter/subscriber/pkg/k8s/log.go
```diff
@@ -96,32 +86,84 @@ func (k8s *k8sSubscriber) CreatePodLog(podLog types.PodLogRequest) (types.PodLog
...
 	}
 
+	return logDetails, nil
+}
+
+// Categorizing log messages by experiment phase
+func (k8s *k8sSubscriber) categorizeLogByPhase(phase string) string {
+	switch phase {
+	case "Completed":
+		return "Experiment pod is deleted"
```
Saranya-jena, line 143: There can be a situation where users haven't added the cleanup chaos step, in that case, the experiment resources won't be deleted. We have to handle that situation as well

kwx4957: @Saranya-jena I think this function will work in that case, because it will get data from mongoDB when getting the pod logs fails, but I will test it for that case.

kwx4957: @Saranya-jena I remove the podGc field and the cleanup chaos step. Is this the correct procedure?

image

Saranya-jena (approved):

amityt (approved):

namkyu1999:
Close this pr because it is no longer maintained

PR #5036, opened 2025-01-31
fix: duplicate name key err for probes

### LINKED ISSUE #5035: Cannot Create Chaos Probe with the Same Name in Different Projects
**What happened**:

Unable to create a chaos probe with the same name in two different projects.

Should be able to create chaos probes with the same name in different projects since project scope should be independent.

Can be corrected in defining the indexes for the chaosProbes collections.

1. Create a chaos probe in one project.
2. Attempt to create another probe with the same name in a different project. An error will be thrown.

Nope

ayush3160:
## Proposed changes

This PR updates the index definition in the `chaosProbes` collection, making the combination of `name` and `project_id` unique instead of `name` field alone.

fixes #5035 

## Types of changes

## Checklist

### DIFF
--- a/chaoscenter/graphql/server/pkg/database/mongodb/init.go
+++ b/chaoscenter/graphql/server/pkg/database/mongodb/init.go
@@ -278,15 +278,11 @@ func (m *MongoClient) initAllCollection() {
 	m.ChaosProbeCollection = m.Database.Collection(Collections[ChaosProbeCollection])
 	_, err = m.ChaosProbeCollection.Indexes().CreateMany(backgroundContext, []mongo.IndexModel{
 		{
-			Keys: bson.M{
-				"name": 1,
+			Keys: bson.D{
+				{Key: "name", Value: 1},
+				{Key: "project_id", Value: 1},
 			},
 			Options: options.Index().SetUnique(true),
-		},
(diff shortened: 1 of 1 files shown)


chaoscenter/graphql/server/pkg/database/mongodb/init.go
```diff
@@ -277,16 +277,12 @@ func (m *MongoClient) initAllCollection() {
...
-			},
-			Options: options.Index().SetUnique(true),
-		},
 		{
 			Keys: bson.D{
-				{"project_id", 1},
+				{Key: "name", Value: 1},
+				{Key: "project_id", Value: 1},
 			},
+			Options: options.Index().SetUnique(true),
```
SarthakJain26, line 285: Slight modification here. This index will also consider the documents that are deleted. So a partial filter expression can be added to only have this check for documents that are not deleted.

```
Options: options.Index().SetUnique(true).SetPartialFilterExpression(bson.D{{
				"is_removed", false,
			}}),
```

SarthakJain26 (changes requested):
Changes looks good to me

PriteshKiri:
Closing the PR due to inactivity.

PR #5321, opened 2025-11-10
Fix bug where passing existing experiment name into UpdateChaosExperiment causes error

### LINKED ISSUE #5320: Calling UpdateChaosExperiment with same name returns error
**What happened**:
When calling the `updateChaosExperiment` mutation and passing in the same experiment name that the experiment already had, an error is thrown, stating that it's a duplicate name.

The mutation should work as usual with no changes to the experiment name.

In `chaoscenter/graphql/server/pkg/chaos_experiment/handler/handler.go`, there is logic to catch duplicate experiment names in `UpdateChaosExperiment`, but it throws an error even if the duplicate experiment name belongs to the original experiment.

Call `updateChaosExperiment` mutation with the experiment name field as the experiment's existing name.

1spyral:
Fixes #5320 

## Proposed changes

Fetches the existing experiments with the declared experiment name and compares it with the experiment being edited. If they are the same, skip the duplicate name check which was previously erroring.

Changes to tests:
- Mock the fetch for the existing experiment
- Rename tests to use `success` and `failure` like other tests in file
- Add test to ensure that passing a duplicate experiment name throws an error

## Types of changes

## Checklist

### DIFF
--- a/chaoscenter/graphql/server/pkg/chaos_experiment/handler/handler.go
+++ b/chaoscenter/graphql/server/pkg/chaos_experiment/handler/handler.go
@@ -283,10 +283,18 @@ func (c *ChaosExperimentHandler) UpdateChaosExperiment(ctx context.Context, requ
 		revID = uuid.New().String()
 	)
 
-	// Check if the experiment_name exists under same project
-	err := c.validateDuplicateExperimentName(ctx, projectID, request.ExperimentName)
+	// Check if the experiment name exists under same project
+	experiments, err := c.chaosExperimentOperator.GetExperiments(bson.D{
+		{"project_id", projectID},
+		{"name", request.ExperimentName},
+	})
 	if err != nil {
 		return nil, err
+	}
+	for _, exp := range experiments {
+		if exp.ExperimentID != *request.ExperimentID {
+			return nil, errors.New("experiment name should be unique, duplicate experiment found with name: " + request.ExperimentName)
+		}
 	}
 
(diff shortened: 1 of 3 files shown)


chaoscenter/graphql/server/pkg/chaos_experiment/handler/handler.go
```diff
@@ -283,12 +283,30 @@ func (c *ChaosExperimentHandler) UpdateChaosExperiment(ctx context.Context, requ
 		revID = uuid.New().String()
 	)
 
-	// Check if the experiment_name exists under same project
-	err := c.validateDuplicateExperimentName(ctx, projectID, request.ExperimentName)
+	// Fetch the existing experiment
```
SarthakJain26: This can further be optimised, by fetching the mongo document by querying it based on name. Then check if the number of documents is more than 1 then
- if request.ExperimentID == wfDetails.ExperimentID which means same document is being modified so the validation can be skipped
- if request.ExperimentID != wfDetails.ExperimentID that means another document exists with the same name so in this case it should throw an error that same name exists in the project.

 With this method you can skip using the validateDuplicateExperimentName and save an extra DB call

1spyral:
@SarthakJain26 thanks for the review! I've optimized the check to use one DB query. Let me know if everything looks good to you!

SarthakJain26 (approved):

PriteshKiri (approved):

1spyral:
Fixed outdated mocks in fuzz test

SarthakJain26:
graphq-server build pipeline failing due to vulnerability scanning. Will fix the vulnerability in another PR.

PR #5323, opened 2025-11-12
fix: include K6 helper pod logs in ChaosCenter backend with proper error handling

### LINKED ISSUE #5208: Issues With Logging When Running K6 Load Test on Litmus
**What happened**:
When I run k6 load tests with litmus, I see "1 mainLogs: Failed to get argo pod logs" in the console log.

I expect to see logs that give me information about the k6 load test run.

I investigated and found k6 experiments create a 4-level k8s pod hierarchy:
1. Argo Workflow Pods: `experiment-name-timestamp-xxxxx` 
2. Chaos Runner: `experiment-runner-xxxxx`
3. Job Pod: `experiment-job-xxxxx`  
4. Helper Pod: `experiment-helper-xxxxx`** ← Contains the actual k6 results

Main issues according to my investigation:
https://github.com/litmuschaos/litmus/blob/master/chaoscenter/subscriber/pkg/k8s/log.go#L56 
This code in log.go does not check for the helper pod and hence doesn't get any of the logs from it back.

The challenge here also seems to be that the logs are fetched from the “live” pods via k8s API calls, so when the pod dies and gets removed the logs are gone hence why we see the “failed to get pod logs” error.

There also seems to be another frontend issue in https://github.com/litmuschaos/litmus/blob/92e8e954c9f242a9695608f8e3f0cbabfaefed3b/chaoscenter/web/src/controllers/CustomStepLog/CustomStepLog.tsx#L69 that causes the underlying “runner” pod logs to also not be displayed (it seems like they are actually returned on the response payload but the JS code does not correctly unwrap it).

I ran a basic k6 load test that just does SELECT 1; on SQL for 30 seconds on a cluster using litmus. Could keep all configurations like number of users to the lowest just to get an end-to-end k6 run.

UJESH2K:
---

### 🧩 What happened  
When running **K6 load tests** using LitmusChaos, users observed the following message in ChaosCenter logs:

Even though the actual K6 test results existed, they were hidden because the backend only fetched logs from the **main** or **runner** pods.  
However, in K6 experiments, the real results are stored in the **helper pod** (`*-helper`).

---

### ⚙️ Root Cause  
K6 experiments follow a 4-level Kubernetes pod hierarchy:

1. **Argo Workflow Pod:** `experiment-name-timestamp-xxxxx`  
2. **Chaos Runner Pod:** `experiment-runner-xxxxx`  
3. **Job Pod:** `experiment-job-xxxxx`  
4. **Helper Pod:** `experiment-helper-xxxxx` ← contains actual K6 load test logs  

The backend code at  
[`chaoscenter/subscriber/pkg/k8s/log.go`](https://github.com/litmuschaos/litmus/blob/master/chaoscenter/subscriber/pkg/k8s/log.go)  
did not include fallback logic to check for the helper pod, leading to empty or failed log retrieval.

---

### 🧠 What this PR does  
✅ Adds **fallback logic** to retrieve logs from the *helper container* when main logs are unavailable.  
✅ Improves **error handling** to clearly report missing log sources.  
✅ Ensures **K6 experiment logs** are now visible in ChaosCenter.  
✅ Keeps existing functionality intact for other experiment types.

**Updated backend logic snippet:**
```go
mainLog, err := k8s.GetLogs(podLog.PodName, podLog.PodNamespace, "main")
if err != nil {
    logrus.Warnf("main container log not found, retrying helper container for pod %v", podLog.PodName)
    helperLog, helperErr := k8s.GetLogs(podLog.PodName, podLog.PodNamespace, "")
    if helperErr != nil {
        logrus.Errorf("failed to get helper container logs for pod %v: %v", podLog.PodName, helperErr)
        logDetails.MainPod = "logs not found"
    } else {
        helperLog = strconv.Quote(strings.Replace(helperLog, `"`, `'`, -1))
        logDetails.MainPod = helperLog[1 : len(helperLog)-1]
    }
} else {
    mainLog = strconv.Quote(strings.Replace(mainLog, `"`, `'`, -1))
    logDetails.MainPod = mainLog[1 : len(mainLog)-1]
}


UJESH2K:
@PriteshKiri I've resolved the error part from the original pr here.

ispeakc0de:
Hi @UJESH2K, the helper pod is a separate pod, not another container within the same pod. So this change won't address the issue. You'll need to implement the same logic for the helper pod, similar to how we fetch the experiment pod logs.

PriteshKiri:
@UJESH2K are you still working on this PR?

UJESH2K:
> @UJESH2K are you still working on this PR?

Hey , I’m not working on this issue. I tried it previously, but my approach wasn’t well suited for this implementation.

PriteshKiri:
Thanks for the update. Closing the issue

PR #5372, opened 2026-01-03
Update sidebar layout & resolve spring-boot build failure

shovan-mondal:
## Proposed changes

This PR fixes a UI layout issue where the sidebar background was getting cut off on long pages.
The sidebar was previously limited to the initial viewport height, causing white gaps when scrolling down.

**Changes:**
- Changed `height: 100vh` to `min-height: 100vh` in `DefaultLayout.module.scss` to allow expansion.
- Added `position: sticky` and `top: 0` to keep the sidebar correctly anchored.
- Added `align-self: stretch` to ensure it fills the parent flex container height.

Fixes #5371
Fixes #5353 

## Types of changes

What types of changes does your code introduce to Litmus? Put an `x` in the boxes that apply
- [ ] New feature (non-breaking change which adds functionality)
- [x] Bugfix (non-breaking change which fixes an issue)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation Update (if none of the other choices applies)

## Checklist

Put an `x` in the boxes that apply. You can also fill these out after creating the PR. If you're unsure about any of them, don't hesitate to ask. We're here to help! This is simply a reminder of what we are going to look for before merging your code.
- [x] I have read the [CONTRIBUTING](https://github.com/litmuschaos/litmus/blob/master/CONTRIBUTING.md) doc
- [x] I have signed the commit for DCO to be passed.
- [x] Lint and unit tests pass locally with my changes
- [x] I have added tests that prove my fix is effective or that my feature works (if appropriate)
- [ ] I have added necessary documentation (if appropriate)

## Dependency
- None

## Special notes:
I have verified this fix locally on Chrome by navigating to long pages (e.g., ChaosHub) and scrolling to the bottom. The sidebar now correctly fills the entire height.

PR #5382, opened 2026-01-07
Resolve experiment run race conditions and persistence state bugs

### LINKED ISSUE #5381: Resolve race conditions and state sync in experiment run
### **What happened:** 
There are multiple related stability and UX issues regarding the Experiment Run lifecycle in the Chaos Studio:

1. **Ghost Loading / Double Submission:** When clicking the "Run" button rapidly, multiple GraphQL mutations are fired before the UI updates to a loading state. This results in duplicate experiment runs spinning up simultaneously, causing resource contention on the cluster.

Image

2. **Stale State on Navigation:** If a user runs an experiment, waits for 5sec then navigates away, and then returns to the Builder/Editor, the UI reverts to showing the "Run" button even though the experiment is still active (it should show "Stop").

https://github.com/user-attachments/assets/2405e550-2b0a-46a7-81d0-27709ed38945

3. **Sidebar & Top Bar Desynchronization:** The "Run" button in the top header and the "Run" icon in the sidebar do not share the same disabled logic. It is possible for one to be enabled while the other is disabled during specific unsavedChanges states.
4. **Sidebar "Stop" Stuck**: If the editor has unsavedChanges=true, the Sidebar fails to poll/refresh the experiment status, causing the "Stop" button to persist even after the run has finished.
5. **No Auto-Refresh:** Sometimes the UI does not automatically revert from "Stop" → "Run" when an active experiment completes; the user is forced to refresh the page.

### **What you expected to happen:**

1. **In-Flight Guard:** The "Run" action should be mutually exclusive. A second click should be blocked immediately (synchronously) if a request is already in flight.
2. **Persistent State:** The UI should correctly reflect the "Running" (Stop button visible) state regardless of navigation history.
3. **Unified Locking:** Both the Top Bar and Sidebar buttons should obey the same rules regarding loading states and unsaved changes.
4. **Independent Polling:** The application should poll for run completion regardless of whether the user has unsaved changes in the editor, and auto-update the UI when the run finishes.

### **How to reproduce it (as minimally and precisely as possible):** 
**Scenario 1 (Ghost Loading):**

- Open one of your chaos Experiment
- In Chaos Studio Builder, rapidly double-click the "Run" button.
- Go to Run History tab
- Observe two runs appearing in the "Run History" tab.

**Scenario 2 (Stale State):**

- Run an experiment.
- While it is running (Stop button visible), click the browser "Back" button, then the "Forward" button, then "back" button.
- Observe the button revert to "Run" despite the experiment still being active.

### **Anything else we need to know?:**

- The "Ghost Loading" issue is critical as it can cause OOM (Out of Memory) kills on smaller clusters by spinning up double the expected chaos runner pods.

shovan-mondal:
### **Proposed changes**

This PR resolves multiple stability and UX issues in the Experiment Execution lifecycle, specifically targeting "Ghost Loading" (double submissions), stale UI states during navigation, and sidebar synchronization.

### **Specific Fixes:**

1. Fixed "Ghost Loading" / Race Condition:
2. Implemented a synchronous useRef guard clause in the execution handler.
3. Prevents duplicate GraphQL mutations when the "Run" button is double-clicked rapidly, protecting the cluster from resource contention.
4. Added visual disabled states to buttons during the loading phase.
5. Fixed Stale State Persistence:
6. Updated the navigation logic to use { replace: true } upon successful run submission.
7. Ensures that navigating "Back" from the details page skips the Builder (Run) page, preventing users from seeing the "Run" button again for an already active experiment.

**Sidebar & Header Synchronization:**

1. Unified the logic for the Top Bar "Run" button and Sidebar "Run" icon. Both now respect the same unsavedChanges + loading + running state derived from the active experiment.
2. Fixed a bug where the Sidebar "Stop" button would get stuck if the editor had unsaved changes (polling logic was previously blocked by the dirty state).

**Auto-Refresh on Completion:**

1. Enabled independent polling for experiment status. The UI now automatically reverts from "Stop" → "Run" when the experiment completes, without requiring a manual page refresh.

Fixes #5381 

### **Types of changes**

What types of changes does your code introduce to Litmus? Put an x in the boxes that apply
[ ] New feature (non-breaking change which adds functionality)
[x] Bugfix (non-breaking change which fixes an issue)
[ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
[ ] Documentation Update (if none of the other choices applies)

Checklist

Put an x in the boxes that apply.
[x] I have read the [CONTRIBUTING](https://github.com/litmuschaos/litmus/blob/master/CONTRIBUTING.md) doc
[x] I have signed the commit for DCO to be passed.
[ ] Lint and unit tests pass locally with my changes
[x] I have added tests that prove my fix is effective or that my feature works (See Proof in linked Issue)
[ ] I have added necessary documentation (if appropriate)

### **Dependency**
This PR depends on the fix for ChaosFaults.tsx (missing isSpringbootFault variable) which is currently pending in my other PR: https://github.com/litmuschaos/litmus/pull/5372

### **Special notes for your reviewer:**

1. **Implementation Detail:** I used a useRef pattern for the "Ghost Loading" fix instead of just useState. This is intentional: standard React state updates are asynchronous and can be too slow to catch rapid double-clicks. The ref provides a synchronous guard to ensure strict mutual exclusion of the mutation.
2. Screenshots: I have attached video recordings & sreenshot demonstrating the race condition and the state persistence behavior in the linked Issue: #5381 .

### DIFF
--- a/chaoscenter/graphql/server/pkg/gitops/git-lock.go
+++ b/chaoscenter/graphql/server/pkg/gitops/git-lock.go
@@ -34,6 +34,7 @@ func (g *GitMutexLock) Unlock(repo string, branch *string) {
 	key := getKey(repo, branch)
 	g.mapMutex.Lock()
 	if _, ok := g.gitMutex[key]; !ok {
+		g.mapMutex.Unlock()
 		return
 	}
 	temp := g.gitMutex[key]
--- a/chaoscenter/graphql/server/pkg/gitops/gitops.go
+++ b/chaoscenter/graphql/server/pkg/gitops/gitops.go
@@ -276,7 +276,7 @@ func (c GitConfig) GitHardReset() error {
 
 // GitPull updates the repository in provided Path
 func (c GitConfig) GitPull() error {
-	_, workTree, err := c.getRepositoryWorktreeReference()
+	repo, workTree, err := c.getRepositoryWorktreeReference()
 	if err != nil {
 		return err
 	}
@@ -284,15 +284,27 @@ func (c GitConfig) GitPull() error {
 	if err != nil {
 		return err
 	}
-	err = workTree.Pull(&git.PullOptions{
-		Auth:          auth,
-		RemoteName:    c.RemoteName,
-		ReferenceName: plumbing.NewBranchReferenceName(c.Branch),
-		SingleBranch:  true,
+	err = repo.Fetch(&git.FetchOptions{
+		Auth:       auth,
+		RemoteName: c.RemoteName,
+		Force:      true,
 	})
 	if err != nil && !errors.Is(err, git.NoErrAlreadyUpToDate) {
 		return err
 	}
+	remoteRefName := plumbing.NewRemoteReferenceName(c.RemoteName, c.Branch)
+	remoteRef, err := repo.Reference(remoteRefName, true)
+	if err != nil {
+		return err
+	}
+	err = workTree.Reset(&git.ResetOptions{
+		Mode:   git.HardReset,
+		Commit: remoteRef.Hash(),
+	})
+	if err != nil {
+		return err
+	}
+
 	return nil
 }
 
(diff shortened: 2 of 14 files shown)


chaoscenter/web/src/components/ExperimentActionButtons/ExperimentActionButtons.tsx
shovan-mondal, line 1: I used a ref here instead of state because the state update was too slow to catch rapid double clicks. This ensures the guard is synchronous.

shovan-mondal:
Apologies for resetting the approval! 😅

I just pushed one final commit adding a regression test (ChaosStudio.test.tsx). I wanted to make sure this race condition is covered by automation so it never comes back.
No changes were made to the actual fix logic. Do re-review it!

PR #5407, opened 2026-01-20
 Fix: Prevent subscriber crash on Workflow ADD events with ChaosEngine nodes

WHOIM1205:
## Summary

This PR fixes a **nil pointer dereference** in the subscriber’s `WorkflowEventHandler` that caused the subscriber pod to **crash on every Workflow ADD event** containing a `ChaosEngine` node.

The issue occurs because informer ADD events pass `oldObj` as `nil`, but the code attempted to dereference `oldObj.Status.Nodes`, resulting in a panic. This led to **CrashLoopBackOff**, **silent loss of experiment status updates**, and **stuck experiments in the UI**.

The fix adds a defensive nil check while preserving the existing UPDATE-event behavior.

---

## Root Cause

- Kubernetes informer ADD events always pass `oldObj = nil`
- `WorkflowEventHandler` assumed `oldObj` was always non-nil
- Code dereferenced `oldObj.Status.Nodes[i]` without a nil guard
- Go evaluates `oldObj.Status` before map access, causing a panic

This happens during:
- Subscriber pod startup (initial informer sync)
- Subscriber restarts (evictions, upgrades)
- Creation of new chaos experiments

---

## Impact Before Fix

- Subscriber crashes immediately on ADD events
- Experiment status updates stop flowing to ChaosCenter
- Chaos experiments appear permanently stuck in the UI
- CI/CD pipelines relying on chaos results fail unpredictably
- Subscriber enters CrashLoopBackOff on restart

---

## Fix Details

- Safely handle ADD events by guarding access to `oldObj`
- Default phase is set from `ChaosData.ExperimentStatus`
- Pending → Running transition logic is applied **only for UPDATE events**
- No behavioral change for UPDATE events
- No refactor or logic duplication

---

## Steps to Reproduce (Before Fix)

1. Deploy Litmus ChaosCenter with a connected infrastructure
2. Run any chaos experiment (e.g., pod-delete)
3. Ensure the workflow contains a `ChaosEngine` node
4. Restart the subscriber pod:
   ```bash
   kubectl delete pod -l app=subscriber -n litmus

### DIFF
--- a/MAINTAINERS.md
+++ b/MAINTAINERS.md
@@ -17,20 +17,20 @@ chaos-sdk        |go/python/ansible sdk           |litmus-go,litmus-python,litmu
 e2e              |e2e-suite, e2e-dashboard         |litmus-e2e                                       |@uditgaurav, @Jonsy13                       |@neelanjan00, @S-ayanide, @avaakash         |
 integrations     |CI/CD plugins, wrappers          |chaos-ci-lib, gitlab-templates, github-actions   |@uditgaurav, @ksatchit                    |@ispeakc0de, @Adarshkumar14         | 
 helm-charts      |control-plane, agent, experiments|litmus-helm                                      |@Jasstkn, @ispeakc0de, @imrajdas, @Jonsy13             |@ksatchit, @uditgaurav         |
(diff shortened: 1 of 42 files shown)


WHOIM1205:
Hi @SarthakJain26 
  

This PR fixes multiple correctness issues in the event-tracker informers (copy-paste bugs affecting UPDATE handling and resource type auditing for StatefulSets and DaemonSets).

Could someone please take a look and review when time permits?  
Happy to address any feedback or add tests if needed. Thanks!

Issue #5413, opened 2026-01-26
Cannot Use Email as Username Due to 16-Character Limit in User Creation API


CorneilleEdi:
**What happened**:  
I want to pre‑create users via the API, using their email as the username, so that users who later log in via Dex can use the already created account through data correlation (the email is used as the username when logging in via Dex: https://github.com/litmuschaos/litmus/blob/master/chaoscenter/authentication/api/handlers/rest/dex_auth_handler.go#L127). However, the API enforces a username constraint: “The username should be at least 3 characters long and at most 16 characters long.” Email addresses are usually longer than 16 characters, so the API rejects those user creations.

**What you expected to happen**:  
I expected to be able to create a user via the API with their full email as the username (even if it’s longer than 16 characters).

**Where can this issue be corrected? (optional)**  
In the user creation/validation logic that enforces the 3–16 character username length limit.

**How to reproduce it (as minimally and precisely as possible)**:

1. Call the user‑creation API with `username` set to a valid email longer than 16 characters, e.g. `john.smith.long@example.com`.  
2. The API returns a 401 validation error, as defined here:  

   https://github.com/litmuschaos/litmus/blob/master/chaoscenter/authentication/pkg/utils/errors.go#L58  
   `The username should be atleast 3 characters long and atmost 16 characters long.`  

**Anything else we need to know?**:  
- There is no functional problem with Dex; the issue is only the username length check in this API.  
- A possible solution would be to relax or make configurable the 16‑character limit.

PriteshKiri:
Hey @DarkBird10020

Thanks for your contribution!

Just a reminder to please start working on an issue only after it has been officially assigned to you. We’re approving the PR this time, but please keep this in mind for future contributions.

@SarthakJain26 has added a comment on your PR, can you please check it once?

PriteshKiri:
This issue is now open for contribution!!

yashgoyal0110:
@PriteshKiri 
assign please

PR #5418, opened 2026-01-28
fix(auth): allow email addresses as usernames in Create User API

DarkBird10020:
## Proposed changes
This PR fixes [Issue #5413](https://github.com/litmuschaos/litmus/issues/5413) where the **Create User API rejects email-based usernames** due to the strict 3–16 character / alphanumeric-only validation. Since Dex login uses the user’s **email as `username`**, pre-creating users via API with their email was not possible.

## What changed
- Updated `ValidateStrictUsername` to **accept email addresses as valid usernames** (up to 255 chars) while **preserving the existing strict rules** for non-email usernames.
- This enables admins to pre-create users using full emails (e.g. `john.smith.long@example.com`) so Dex login can match the pre-created account.

## Why this is safe
- **Backwards compatible**: non-email usernames still follow the original strict validation.
- Only affects username validation at user creation/update entrypoint(s) that call this validator.
- Does not modify Dex flow; it just aligns API validation with how Dex already populates usernames.

## Files changed
- `chaoscenter/authentication/pkg/utils/sanitizers.go`

## Test plan
- Create a user via API with `username = john.smith.long@example.com` → **should succeed**
- Create a user with an invalid email username (e.g. `user@`) → **should fail**
- Create a user with a regular username within 3–16 chars (e.g. `test_user`) → **should still succeed**
- Login via Dex with the same email → **should match the pre-created user account**

## Type of change


SarthakJain26:
@DarkBird10020 can you please confirm what happens when an user has created a account using dex and then used the same email as username to create another account this time without dex.

PriteshKiri:
@DarkBird10020 any updates??

PriteshKiri:
Closing due to inactivity

Issue #5434, opened 2026-02-09
Experiment manifest contains literal "undefined" image prefix when project has no Image Registry configured


KelvinVenancio:
**What happened**:

When creating an experiment via the Chaos Center UI with GitOps enabled, the manifest saved to the Git repository contains broken image references with a literal `undefined` prefix:

```yaml
# install-chaos-faults template - go-runner image
image: undefined/go-runner:3.24.0

# fault engine template - litmus-checker image
image: undefined/litmus-checker:2.11.0
```

This causes `ImagePullBackOff` when the experiment is executed, as Kubernetes tries to pull from a non-existent `undefined/` registry.

The same experiment created via UI **without** GitOps produces the correct images:

```yaml
image: litmuschaos.docker.scarf.sh/litmuschaos/go-runner:3.24.0
image: docker.io/litmuschaos/litmus-checker:2.11.0
```

**What you expected to happen**:

The manifest should contain valid image references regardless of whether Image Registry is explicitly configured in the project settings.

**Where can this issue be corrected?**

The bug is in the `updateContainerImage` function at `chaoscenter/web/src/services/experiment/KubernetesYamlService.ts` (line 1174):

```typescript
function updateContainerImage(image: string, registry: ImageRegistry | undefined): string {
  const parts = image.split('/');
  const lastPart = parts[parts.length - 1];
  const newImage = `${registry?.repo}/${lastPart}`;  // <-- produces "undefined/go-runner:3.24.0"
  return newImage;
}
```

The optional chaining (`registry?.repo`) prevents a runtime error, but JavaScript template literals stringify `undefined` to the literal string `"undefined"`.

The data flow is:

1. `StudioOverview.tsx` fetches the image registry via `getImageRegistry({ projectID })`
2. If the project has **no Image Registry configured** in Settings → Image Registry, the query returns null
3. `imageRegistry` is set to `undefined` and stored on the experiment object in IndexedDB
4. When a fault is added/updated, `addFaultToManifest` / `updateFaultInManifest` call `updateContainerImage` with `experiment.imageRegistry` which is `undefined`
5. The template literal produces `"undefined/go-runner:3.24.0"`

Other code paths in the same file already handle this case correctly:

```typescript
// Line 1122 - preProcessChaosEngineManifest (safe: nullish coalescing)
const imageRegistry = experiment?.imageRegistry?.repo ?? 'litmuschaos';

// blankCanvasTemplate.ts (safe: default parameter)
[...]

PR #5435, opened 2026-02-09
fix(web): handle undefined image registry in updateContainerImage

KelvinVenancio:
## Proposed changes

Fixes https://github.com/litmuschaos/litmus/issues/5434

When a project has no Image Registry explicitly configured in Settings → Image Registry, the `updateContainerImage` function produces broken image references with a literal `undefined` prefix (e.g., `undefined/go-runner:3.24.0`), causing `ImagePullBackOff` on experiment execution.

The fix adds an early return when `registry` is not configured, preserving the original image string which already contains the correct registry path from the ChaosHub fault definition:

```diff
 function updateContainerImage(image: string, registry: ImageRegistry | undefined): string {
+  if (!registry?.repo) return image;
+
   const parts = image.split('/');
   const lastPart = parts[parts.length - 1];
-  const newImage = `${registry?.repo}/${lastPart}`;
+  const newImage = `${registry.repo}/${lastPart}`;
   return newImage;
 }
```

This is consistent with how other code paths in the same file handle undefined registry:
- `preProcessChaosEngineManifest` (line 1122): `experiment?.imageRegistry?.repo ?? 'litmuschaos'`
- `blankCanvasTemplate.ts`: default parameter `{ repo: 'litmuschaos', secret: '' }`

## Types of changes

## Checklist

## Dependency

None.

- Single line fix — adds an early return guard for undefined registry.
- The optional chaining `?.` was already present but only prevents runtime errors; it does not prevent template literals from stringifying `undefined` to the literal string `"undefined"`.
- The removed `?.` in `registry.repo` on the template literal is safe because the early return guarantees `registry` is defined at that point.
- No existing unit tests for `KubernetesYamlService.ts` in the repository.

### DIFF
--- a/chaoscenter/web/src/services/experiment/KubernetesYamlService.ts
+++ b/chaoscenter/web/src/services/experiment/KubernetesYamlService.ts
@@ -1172,13 +1172,15 @@ export class KubernetesYamlService extends ExperimentYamlService {
 }
 
 function updateContainerImage(image: string, registry: ImageRegistry | undefined): string {
+  if (!registry?.repo) return image;
+
   // Extract image name + tag from original image
   // Example: docker.io/litmuschaos/litmus-checker:2.11.0
   const parts = image.split('/');
   const lastPart = parts[parts.length - 1]; // litmus-checker:2.11.0
 
   // Build new image
-  const newImage = `${registry?.repo}/${lastPart}`;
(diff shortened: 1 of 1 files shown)


KelvinVenancio:
Closing this PR — after further testing, the issue could not be reproduced. It was likely caused by a transient browser state rather than a code defect.

PR #5470, opened 2026-04-01
fix: allow stopping experiments when infra is disconnected

mugiwaraluffy56:
## Proposed changes

  Fixes #5466

when an infra agent is redeployed it gets a new connection ID, so any experiments that were already running get stuck in `Running` state forever. the stop signal just gets silently dropped since the old infra channel no longer exists and the UI shows "INFRA ERROR: INFRA NOT CONNECTED" with no way to cancel or stop anything.

the problem was that `ProcessExperimentRunStop` only touched `updated_at` and `updated_by` in the DB, fully relying on the agent to report back and flip the phase to `Stopped`. that callback never comes if the infra is disconnected. on top of that, `StopExperimentRuns` was passing a broad experiment-level query without scoping it to the specific run ID, which would have caused wrong updates once the DB write started carrying real fields.

the fix makes `ProcessExperimentRunStop` write `phase = "Stopped"` and `completed = true` directly to the DB so the run is always marked stopped from the server side regardless of infra state. the subscriber notification is still sent as a best-effort in case the infra is connected and needs to clean up. `StopExperimentRuns` now also builds a per-run query with
  `experiment_run_id` to scope the update correctly.

  ## Types of changes

  ## Checklist

  ## Dependency
  - None

  no schema changes, no new fields. if infra is connected it still gets the stop signal as before. if not, the DB state is correct regardless.


mugiwaraluffy56:
hey @Jonsy13 @imrajdas @gdsoumya could you review this when you get a chance?

PriteshKiri:
Hey @mugiwaraluffy56 

@AmitKumarDas will look into it soon.

amityt (approved):

amityt:
This was a long pending issue. Thanks for the fix @mugiwaraluffy56  🚀

mugiwaraluffy56:
thanks :)

Jonsy13 (approved):

PR #5493, opened 2026-04-30
fix(chaoscenter): resolve blank page for CronWorkflow run history

xorjnox:
## issue 
https://github.com/litmuschaos/litmus/issues/5494

## Proposed changes

This PR fixes a bug where the CronWorkflow Run History page renders as a blank page.
**Root cause**: In RightSideBarV2.tsx, the listExperiment function (which internally calls Apollo's useQuery hook) was being invoked conditionally inside an if (experimentType ===ExperimentType.CRON) block. This violates React's Rules of Hooks — hooks must be called unconditionally on every render in the same order.

The bug manifests specifically for CronWorkflow experiments because:
  1. On the first render, experimentType is undefined(data still loading) → the if is false → hook is not called → React registers   0 hook calls.
  2. On the second render, experimentType resolves to CRON → the if is true → hook is called → React
  detects a new hook appearing mid-render and throws Minified React error `#311` ("Cannot update a component while rendering a different component"),resulting in a blank page.

For NON_CRON experiments, experimentType resolves to NON_CRON so the if stays false on both renders — no mismatch, no crash.

### Before
Screenshot 2026-04-30 at 9 03 48 AM
 

### After 
Screenshot 2026-04-30 at 9 01 32 AM

> Also excludes the auto-generated YamlBuilder.module.scss.d.ts from Codacy analysis via      codacy.yaml. This file is generated by webpack's CSS modules typing and uses declare  namespace which is required by the generator — removing the file causes web-unit-tests to fail.       

                                                                                
## Types of changes

## Checklist

## Dependency


PriteshKiri:
@Ranjan-S1227 could you please resolve the conflicts?

xorjnox:
Resolved @PriteshKiri !

SarthakJain26 (approved):

PriteshKiri (approved):

PR #5519, opened 2026-05-21
Fix/add probe uniqueness check

### LINKED ISSUE #5518: Server-side uniqueness validation missing in AddProbe
## What happened:

`AddProbe` in `pkg/probe/handler/handler.go` has no server-side uniqueness validation before inserting a new probe into the database. A TODO comment explicitly marks this as unresolved:
```
// TODO: Add check if probe exists
```
Uniqueness is currently only enforced at two layers:
- UI layer — validateUniqueProbe is called in the wizard before advancing steps
- MongoDB — a partial compound index on (name, project_id) where is_removed: false prevents duplicate inserts at the storage level

This leaves non-UI callers completely unhandled.

## What you expected to happen:

Calling addProbe with a duplicate probe name via the GraphQL API directly should return a clear, structured application-level error — not a raw MongoDB E11000 duplicate key error.

## Where can this issue be corrected? (optional):

`chaoscenter/graphql/server/pkg/probe/handler/handler.go` : Inside AddProbe, before the `CreateProbe` call. The uniqueness logic already exists as `ValidateUniqueProbe `in the same file and can be reused directly.

## How to reproduce it (as minimally and precisely as possible):

1. Create a probe via the GraphQL API or UI with name test-probe-1 in project project-1
2. Call addProbe again directly via the GraphQL API (bypassing the UI wizard) with the same name and project:
3. Observe: raw MongoDB E11000 duplicate key error is returned instead of a clear application error
4. Same behavior occurs when AddProbe is called directly during experiment import via chaos_experiment/ops/service.go

Image

## Anything else we need to know?:

- This is inconsistent with how other resources handle duplicates: Experiments, infra, and chaos hub already validate duplicates at the handler level (validateDuplicateExperimentName, infra registration check)
- The ValidateUniqueProbe method already exists and is used by the UI: The fix is wiring it into AddProbe, not writing new query logic

yoonseo-han:
## Proposed changes

Fixes #5518 

`AddProbe` missed server-side uniqueness enforcement. Uniqueness was only guaranteed by the MongoDB partial compound index on (name, project_id) where is_removed: false, which prevented data corruption but surfaced raw E11000 duplicate key errors to non-UI callers instead of a clear application-level message.

Two entry points were affected:
- Direct GraphQL/API clients calling addProbe with a duplicate name received an opaque Mongo error
- Experiment import via chaos_experiment/ops/service.go calls AddProbe directly with no duplicate handling at all

Contribution points: 
- This PR wires the existing `ValidateUniqueProbe` method into `AddProbe` as an early return before any database write is attempted, making all entry points consistent.
- Introduces the first unit tests for the probe handler layer: Previously `pkg/probe/handler/` had no test coverage covering both the duplicate rejection behavior and the ValidateUniqueProbe method directly.

## Types of changes

### Changes

`pkg/probe/handler/handler.go`: Added uniqueness check at the top of `AddProbe` using the existing `ValidateUniqueProbe` method, returning a clear "probe already exists" error before any object construction or DB write occurs. 

`ValidateUniqueProbe` runs a `CountDocuments` query filtered on `name`, `project_id`, and `is_removed`= false : Filter is backed by the existing partial compound index on (name, project_id), so no collection scan occurs. 
- If the count is greater than zero the probe name is already active in that project and the request is rejected early. 
- Soft-deleted probes (is_removed: true) are excluded from the count, preserving the existing behavior that allows name reuse after deletion.

`pkg/probe/handler/handler_test.go` — Added unit tests covering:
- Duplicate probe name rejection (`TestAddProbe_DuplicateName`): verifies Create is never called on duplicate
- `ValidateUniqueProbe` returns true when name is unique (`TestValidateUniqueProbe_Unique`)
- `ValidateUniqueProbe`returns false when duplicate exists (`TestValidateUniqueProbe_Duplicate`)
- `ValidateUniqueProbe` propagates DB errors correctly (`TestValidateUniqueProbe_DBError`)

## Checklist

## Dependency

NA

- The MongoDB partial unique index on (name, project_id) where is_removed: false remains the hard guarantee against duplicate data. This PR adds application-level enforcement on top of it, not instead of it. 
- The current approach was chosen for consistency with how infra registration and experiment naming handle the same problem in this codebase.
- There are currently no other probe handler tests. The four tests added here are the first coverage of this layer.


ispeakc0de (approved):

yoonseo-han:
Hi @ispeakc0de @PriteshKiri 

Thanks for reviewing this PR. I have read through the comments that copilot has left. I believe the comments are pointing out a real issue in the shared mongo mock implementation where methods are currently defined with value receivers instead of pointer receivers. 

This can make testify call-history based assertions unreliable when the mock is used via the mongodb.MongoOperator interface. So the issue raised by Copilot seems valid as a test-infra issue.

For this PR however, the probe handler test intent looks correct for validating unique probe and duplicate handling. The root cause appears to be in the shared mock type rather than the probe handler test logic itself.

Hence I wanted to confirm the preferred scope:
1. I can include a focused fix here to update mongo mock methods to pointer receivers,
2. I can keep this PR scoped to probe changes and open a separate issue/PR for the shared mongo mock fix.

Or any other ideas would be very much appreciated. 

Thanks

ispeakc0de:
Hi @yoonseo-han , 
Looks like a straightforward change, so it's better to address it in this PR itself.

yoonseo-han:
Thanks @ispeakc0de 

Included the mongo mock fix in this PR. Updated the MongoOperator mock methods to pointer receivers so that testify assertions work correctly 

Ready for another look when you are free

PR #5521, opened 2026-05-23
Fix ineffective fuzz test assertions in `chaoscenter/subscriber/pkg/events/utils_fuzz_test.go`.

dongjune8931:
## Proposed changes

Fix ineffective fuzz test assertions in `chaoscenter/subscriber/pkg/events/utils_fuzz_test.go`.
While running the fuzz tests locally, I noticed that all three fuzz tests always pass regardless of input, which means they are not actually validating anything.

**FuzzGetExperimentStatus:**
  - `assert.Equal(t, "status is invalid", "status is invalid")` compares two identical string literals — this always passes regardless of what `err` contains.
  - Fixed by using `assert.EqualError(t, err, "status is invalid")` to validate the actual error message.

  **FuzzStrConvTime:**
  - Used `t.Log()` instead of assertions — the test can never fail.
  - Fixed by adding proper assertions that verify return values against expected output.

  **FuzzGenerateWorkflowPayload:**
  - Struct fields were unexported (`cid`, `accessKey`, etc.), so `go-fuzz-headers` `GenerateStruct` cannot populate them via reflection — fuzzing always runs with zero values.
  - Fixed by exporting struct fields (`Cid`, `AccessKey`, etc.).
  - Removed `t.Errorf` on error return, as fuzz tests should only catch panics, not expected errors from random input.

Also removed unused `gomock` dependency from all three fuzz functions.

## Types of changes

## Checklist

## Dependency

Before fix, `FuzzGetExperimentStatus` with `-fuzztime=5s` reported `new interesting: 0` (no new code paths explored). After fix, it reports `new interesting: 16`, confirming that the fuzzer is now actually generating diverse inputs and exploring multiple code paths.


PriteshKiri:
Hey @dongjune8931 
Could you please check the review comments by the Co-Pilot?

dongjune8931:
@PriteshKiri Thanks for the heads-up! I've addressed both Copilot comments:
- Moved operator initialization outside the fuzz loop in `FuzzGenerateWorkflowPayload` to avoid unnecessary allocations per iteration.
- Added both-branch assertion in `FuzzGetExperimentStatus` — now validates `EqualError` when `Phase == ""` and `NoError` otherwise.

All three fuzz tests pass locally.

SarthakJain26 (approved):

ispeakc0de (approved):

PR #5524, opened 2026-05-26
Fix CronWorkflow CR not being updated on target ChaosInfrastructure

xorjnox:
```
The docker-build-graphql-server Trivy scan failure is pre-existing and unrelated to these changes. The vulnerabilities reported (krb5-libs, libcap, go-billy/v5, go-git/v5) exist in the current master build and are not introduced by this PR.     
  ```

 ## Proposed changes                                                                                                         
Fix CronWorkflow CR not being updated on target ChaosInfrastructure when a ChaosExperiment is edited in ChaosCenter.     
                                                                                                                           
 ###  Root cause
SaveChaosExperiment in handler.go called ProcessExperimentUpdate with a hardcoded nil for StateData. In    ops/service.go, the subscriber dispatch (SendExperimentToSubscriber) is gated on r != nil — so every save/edit via the UI silently updated MongoDB but never sent any message to the in-cluster subscriber. The CronWorkflow CR remained stale   until manually deleted.                                                                                                  
                                                                                                                           
###  Fix                                                                                                                   
  - Added r *store.StateData parameter to SaveChaosExperiment and passed data_store.Store from the resolver, matching the pattern already used by UpdateChaosExperiment and DeleteChaosExperiment.                                                 
  - For CronWorkflow experiments: instead of a plain update, the handler now sends a delete of the old CR followed by a  create with the new manifest. This guarantees the Argo CronWorkflow controller picks up all changes (schedule,           
  spec.workflowSpec, metadata). The delete payload uses the previous revision's manifest to ensure the correct CR name is targeted even if the experiment was renamed.                                                                             
  - For non-cron experiments: ProcessExperimentUpdate now correctly dispatches an update action to the subscriber (was     previously a no-op due to nil).                                                                                          
  - First-time saves (ProcessExperimentCreation) also now pass r so the CR is created immediately on active infra.         
                                                                                                                                                                                                                                   
                                                                             
  ## Types of changes                                                                                                      
                                                                                                                         
  ## Checklist       

  ## Dependency                                                                                                               
None.                                                                                                                    
                                                                                                                         
  The subscriber already supports delete and create request types (subscriber/pkg/k8s/operations.go) — no subscriber       
  changes were needed. The subscriber's delete path returns nil on IsNotFound, so a missing CR (e.g. infra was offline   
  during original creation) is handled gracefully. If the infra is inactive, SendRequestToSubscriber is a no-op (no        
  observer in ConnectedInfra), which is consistent with existing behaviour across all other mutations.                   

  The StateData (r) parameter is nil in unit tests and the fuzz test — this is intentional and matches how all other       
  handler tests are written (subscriber dispatch is not tested at the handler layer).

### DIFF
--- a/chaoscenter/graphql/server/graph/chaos_experiment.resolvers.go
+++ b/chaoscenter/graphql/server/graph/chaos_experiment.resolvers.go
@@ -95,7 +95,7 @@ func (r *mutationResolver) SaveChaosExperiment(ctx context.Context, request mode
 		return "", err
 	}
 
-	uiResponse, err = r.chaosExperimentHandler.SaveChaosExperiment(ctx, request, projectID, username)
+	uiResponse, err = r.chaosExperimentHandler.SaveChaosExperiment(ctx, request, projectID, data_store.Store, username)
 	if err != nil {
 		logrus.WithFields(logFields).Error(err)
 		return "", err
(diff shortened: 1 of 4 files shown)


PriteshKiri:
@xorjnox are you still wokring  on this issue?

PriteshKiri:
Closing due to inactivity

PR #5561, opened 2026-06-16
fix(auth): return error when bcrypt fails in CreateUser handler

### LINKED ISSUE #5552: [Security]bcrypt error silently ignored in CreateUser — users created with invalid password hash
## Expected behavior

When `bcrypt.GenerateFromPassword()` fails during 
user creation, the error should be returned to the 
caller and user creation should be aborted.

## Actual behavior

In `chaoscenter/authentication/api/handlers/rest/user_handlers.go` 
line 81, the error from `bcrypt.GenerateFromPassword()` 
is logged but NOT returned. Execution continues with 
an empty/zero-value password hash, meaning users can 
be created with invalid password hashes.

This is a security vulnerability — a user account 
created with an empty hash could potentially be 
accessed without a valid password.

## Location

- File: `chaoscenter/authentication/api/handlers/rest/user_handlers.go`
- Line: 81

## Current Code

hash, err := bcrypt.GenerateFromPassword(...)
if err != nil {
    log.Error(err)
}
// execution continues with empty hash

## Proposed Fix

hash, err := bcrypt.GenerateFromPassword(...)
if err != nil {
    log.Error(err)
    return err
}

## Why This Matters

- Security bug — invalid password hashes compromise
  authentication integrity
- User accounts created during bcrypt failures are
  in an inconsistent state
- Silent failures make debugging extremely difficult

## Additional Notes

Found while auditing the codebase as a new
contributor. I would like to work on this fix
if maintainers agree with the approach.

## Please complete the following information

- File: chaoscenter/authentication/api/handlers/rest/user_handlers.go
- Language: Go
- Component: Authentication service
- Litmus release: latest main branch

g-k-s-03:
Fixes a security bug where bcrypt.GenerateFromPassword() failure was silently ignored in the CreateUser handler. When bcrypt failed, the error was logged but execution continued with an empty password hash, causing a user account to be created in an invalid state with no valid password — while returning 200 OK to the caller.

This PR adds the missing HTTP error response and return statement so the handler aborts correctly on bcrypt failure, consistent with all other error blocks in the same file.

Fixes #5552 

For the checklist, check these boxes:

Bugfix
I have read the CONTRIBUTING doc
I have signed the commit for DCO to be passed
For Special notes for your reviewer, paste this:

The fix is minimal — exactly 2 lines added inside the existing if err != nil block at line 82 of user_handlers.go. No other logic was modified. The error response uses ErrServerError which is the correct code for a server-side bcrypt failure, matching the pattern used throughout the rest of the file.


g-k-s-03:
Hi @PriteshKiri I've addressed both Copilot review comments. Could you please approve the workflow run so CI can execute? Happy to make any further changes

ispeakc0de (approved):

PR #5563, opened 2026-06-18
fix(subscriber): handle all errors in IsAgentConfirmed to prevent nil pointer panic

### LINKED ISSUE #5559: IsAgentConfirmed panics with nil pointer dereference on transient Kubernetes API errors
What happened:
In chaoscenter/subscriber/pkg/k8s/operations.go, the function IsAgentConfirmed() (lines 136–154) can crash with a nil pointer panic.

It calls clientset.CoreV1().ConfigMaps(InfraNamespace).Get(...) and only checks for a "not found" error using k8s_errors.IsNotFound(err). But if Get fails for any other reason — like a temporary API server timeout, network issue, or missing RBAC permissions — getCM comes back as nil. The code then still runs getCM.Data["IS_INFRA_CONFIRMED"], which tries to read from a nil object and crashes the subscriber.

There's also a small piece of dead code right after:

getSecret, err := clientset.CoreV1().Secrets(InfraNamespace).Get(...)
if err != nil {
    return false, "", errors.New(InfraSecretName + " secret not found")
}

if k8s_errors.IsNotFound(err) {   // this can never run — err is already nil here
    return false, "", err
}
Since the line above already returns on any error, this IsNotFound check can never be reached.

What you expected to happen:
Any error from the ConfigMap Get call — not just "not found" — should be handled before getCM is used, so a temporary Kubernetes API hiccup returns an error instead of crashing the agent. The leftover dead IsNotFound check on the secret fetch should also be cleaned up.

Where can this issue be corrected? (optional):
chaoscenter/subscriber/pkg/k8s/operations.go → function IsAgentConfirmed → lines 136–154.

How to reproduce it (as minimally and precisely as possible):

Run the subscriber agent against a cluster.
Make the ConfigMaps(InfraNamespace).Get(InfraConfigName, ...) call fail with something other than "not found" — for example, remove the agent's permission to read configmaps in InfraNamespace, or simulate an API server timeout.
IsAgentConfirmed() panics with a nil pointer dereference on getCM.Data["IS_INFRA_CONFIRMED"] instead of returning the error, crashing the subscriber.
Anything else we need to know?:
Suggested fix — check err != nil first so getCM is never touched when it's nil, and drop the dead error check on the secret fetch:

getCM, err := clientset.CoreV1().ConfigMaps(InfraNamespace).Get(context.TODO(), InfraConfigName, metav1.GetOptions{})
if err != nil {
    if k8s_errors.IsNotFound(err) {
        return false, "", errors.New(InfraConfigName + " configmap not found")
    }
    return false, "", err
}

if getCM.Data["IS_INFRA_CONFIRMED"] == "true" {
    getSecret, err := clientset.CoreV1().Secrets(InfraNamespace).Get(context.TODO(), InfraSecretName, metav1.GetOptions{})
    if err != nil {
        return false, "", fmt.Errorf("failed to get %s secret: %w", InfraSecretName, err)
    }
    return true, string(getSecret.Data["ACCESS_KEY"]), nil
}

return false, "", nil
This is a bug (causes a runtime panic), severity High — it can take down the subscriber on a routine periodic check whenever the Kubernetes API has a transient issue.

AruneshDwivedi:
IsAgentConfirmed panicked on any K8s API error other than NotFound (nil pointer on response body). Fixed by returning early on any ConfigMap Get error (not just checking the body). Extracted the core logic into isAgentConfirmedWithClientset for testability with fake client, covering NotFound, connection errors, and Secret fetch failures. Added unit tests for all three error cases.

Closes #5559


AruneshDwivedi:
The CI failure on this PR is from the docker-build-subscriber step, which runs a Trivy vulnerability scan on the Docker image. The scan found CVE-2026-42506 in the base image dependencies — this isn't related to our code change. The actual tests pass. Subscriber unit test (TestSubscriber_handleAgentConfirmedErrors) passes. Is there a specific action you'd like us to take regarding the Trivy scan result?

PriteshKiri:
@AruneshDwivedi  could you please address the review comments?

AruneshDwivedi:
Addressed. Extracted the core logic into isAgentConfirmedWithClientset so it accepts a kubernetes.Interface — added unit tests covering NotFound, non-NotFound (e.g. connection refused), and Secret fetch errors. PR description already matches what the code does: return early on non-NotFound ConfigMap errors.

ispeakc0de (approved):

AruneshDwivedi:
Done. Added unit tests for isAgentConfirmedWithClientset covering the three cases Copilot flagged: ConfigMap NotFound, a non-NotFound error (returns early, no panic), and a Secret fetch failure (wrapped with context). The PR description now matches the implementation: it returns early on any ConfigMap Get error and wraps the secret error, rather than describing a body/status-code nil-check that was not the actual fix.

AruneshDwivedi (commented):
.

PR #5583, opened 2026-07-16
fix: default image registry repo to prevent undefined image paths on …

### LINKED ISSUE #5510: bug: Experiment cannot run after editing it, due to undefined image path
**What happened**:

When editing an existing chaos experiment and modifying a fault in the Experiment Builder tab, the container image paths in the generated YAML are corrupted. For example, `docker.io/litmuschaos/go-runner:3.27.0` becomes `undefined/go-runner:3.27.0`. **This causes the experiment to fail at runtime because the image cannot be pulled.**

This has been observed on the 3.27.0 version.

The image path should always use the project's configured image registry (defaulting to `docker.io/litmuschaos`) regardless of which tab the user visits first.

The imageRegistry field is only set on the experiment object in IndexedDB when the user submits the Overview tab form (StudioOverview.tsx). The Edit flow (ChaosStudioEdit.tsx) populates the experiment in IndexedDB without imageRegistry, so if the user goes directly to the Builder tab, the field is missing.

The loading of the imageRegistry needs to happen for Experiment Builder tab as well. It is also probably better to default to `docker.io/litmuschaos` rather than having it undefined.

1. Clear browser IndexedDB (DevTools → Application → IndexedDB → delete chaos-db), or use a fresh/incognito browser session
1. Open an existing chaos experiment in Edit mode. You will end up on the Experiment Builder tab.
1. Make a change to a fault, for example make a change to the duration, and save.
1. Inspect the generated YAML — image paths will show undefined/go-runner:3.27.0 instead of docker.io/litmuschaos/go-runner:3.27.0

A workaround to this problem is to always go to the Overview tab before editing an experiment.

yashgoyal0110:
## Proposed changes

fixes #5510

When a chaos experiment is edited and a fault is modified directly from the
**Experiment Builder** tab (without visiting the **Overview** tab), the generated
manifest produced broken image paths such as `undefined/go-runner:3.27.0`, causing
image pull failures at runtime.

The `imageRegistry` is only persisted to IndexedDB when the Overview form is
submitted. The `updateContainerImage` helper in `KubernetesYamlService.ts` — the
single place all fault/engine images are rebuilt — interpolated `registry?.repo`
directly, so a missing registry resolved to the literal string `undefined`.

This change adds a fallback to the default `litmuschaos` repo when the registry is
missing, matching the convention already used elsewhere in the same file
(`experiment?.imageRegistry?.repo ?? 'litmuschaos'`):

const repo = registry?.repo || 'litmuschaos';
const newImage = `${repo}/${lastPart}`;

## Checklist


yashgoyal0110:
@PriteshKiri 
kindly review!

ispeakc0de (approved):

PR #5584, opened 2026-07-17
fix(web): serialize ChaosEngine tolerations as arrays

### LINKED ISSUE #5564: Bug: Advanced Options tolerations serialized as object instead of array in ChaosEngine, save fails with "failed to unmarshal chaosengine"
**What happened**:

When configuring **tolerations** via **Advanced Options** in the ChaosCenter UI (Kubernetes chaos scenario), saving the experiment fails with `failed to unmarshal chaosengine`.

Inspecting the generated workflow YAML shows that `tolerations` under `ChaosEngine.spec.components.runner` and `ChaosEngine.spec.experiments[].spec.components` are written as a single object instead of an array, which does not match the Kubernetes / chaos-operator API schema.

**Incorrect (generated by UI):**
```yaml
tolerations:
  effect: NoExecute
  key: dedicated
  operator: Equal
  value: my-node-tag
```

Scenario saves successfully. The ChaosEngine manifest contains tolerations as an array under `ChaosEngine.spec.components.runner` and `ChaosEngine.spec.experiments[].spec.components`.

**Correct syntax:**
```yaml
tolerations:
  - effect: NoExecute
    key: dedicated
    operator: Equal
    value: my-node-tag
```

chaoscenter/web — ChaosCenter frontend (experiment Advanced Options → ChaosEngine YAML generation)

Primary fix:
- [chaoscenter/web/src/services/experiment/KubernetesYamlService.ts](https://github.com/litmuschaos/litmus/blob/master/chaoscenter/web/src/services/experiment/KubernetesYamlService.ts)
  - wrap tolerations in an array `(tolerations: [tolerations])` for `manifest.spec.components` and `manifest.spec.experiments[0].spec.components` in `updateTolerationInChaosEngine()`

Related type fixes:
- [chaoscenter/web/src/models/chaosEngine.ts](https://github.com/litmuschaos/litmus/blob/master/chaoscenter/web/src/models/chaosEngine.ts#L106)
  - `Runner.tolerations` and `FaultComponents.tolerations` should be `kubernetes.Toleration[]`

1. Open ChaosCenter → create or edit a Kubernetes chaos scenario with at least one fault (e.g. `pod-http-status-code`).
2. Open Advanced Options for the fault.
3. Enable Tolerations and set values, for example:
     - key: `dedicated`
     - value: `my-node-tag`
     - effect: `NoExecute`
     - operator: `Equal`
4. Click **Save**.
As a result, save fails with "failed to unmarshal chaosengine".

Verify in YAML (optional): Export or inspect the workflow manifest. At Workflow.spec tolerations are correct (array):
```yaml
tolerations:
  - effect: NoExecute
    key: dedicated
    ...
```
But in the ChaosEngine artifact they are wrong (object):
```yaml
spec:
  experiments:
    - spec:
        components:
          tolerations:
            effect: NoExecute   # should be a list item under `-`
            key: dedicated
```
**Workaround**: After adding tolerations in Advanced Options, edit the generated YAML manually to use the correct array syntax.
**Environment**: Litmus 3.30.0

UJESH2K:
## Summary

This PR fixes the way tolerations are generated in the ChaosEngine manifest.

Previously, the frontend generated `tolerations` as a single object instead of an array. Since Kubernetes expects an array, saving the ChaosEngine failed with a `failed to unmarshal chaosengine` error.

## Changes

- Serialize runner tolerations as an array.
- Serialize experiment tolerations as an array.
- Update the TypeScript models to use `kubernetes.Toleration[]`.
- Add a regression test to verify that runner and experiment tolerations are serialized as arrays in the generated ChaosEngine manifest.

## Verification

- Verified the generated ChaosEngine manifest now contains `tolerations` as arrays.
- Added a regression test covering the toleration serialization path to prevent future regressions.
- Ran `yarn typecheck` successfully.
- Ran `yarn test` successfully (including the new regression test).

## Note

The repository's pre-commit hook (`lint-staged`) automatically reformatted part of `KubernetesYamlService.ts`. The additional changes are formatting only and do not change the behavior of the code.

Fixes #5564

### DIFF
--- a/chaoscenter/web/src/models/chaosEngine.ts
+++ b/chaoscenter/web/src/models/chaosEngine.ts
@@ -103,7 +103,7 @@ export interface RunnerInfo {
   // Secrets for runner pod
   secrets?: kubernetes.Secret[];
   // Tolerations for runner pod
-  tolerations?: kubernetes.Toleration;
+  tolerations?: kubernetes.Toleration[];
   // Resource requirements for the runner pod
   resources?: kubernetes.ResourceRequirements;
 }
@@ -341,7 +341,7 @@ export interface FaultComponents {
   nodeSelector?: { [key: string]: string };
   statusCheckTimeouts?: StatusCheckTimeout;
   resources?: kubernetes.ResourceRequirements;
-  tolerations?: kubernetes.Toleration;
+  tolerations?: kubernetes.Toleration[];
 }
 
 // StatusCheckTimeout contains Delay and timeouts for the status checks
--- a/chaoscenter/web/src/services/experiment/KubernetesYamlService.ts
+++ b/chaoscenter/web/src/services/experiment/KubernetesYamlService.ts
@@ -455,12 +455,12 @@ export class KubernetesYamlService extends ExperimentYamlService {
       ...manifest.spec.components,
       runner: {
         ...manifest.spec.components?.runner,
-        tolerations: tolerations
+        tolerations: [tolerations]
       }
     };
     manifest.spec.experiments[0].spec.components = {
       ...manifest.spec.experiments[0].spec.components,
-      tolerations: tolerations
+      tolerations: [tolerations]
     };
 
     return manifest;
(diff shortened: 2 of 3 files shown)


Aryanbhargava18 (commented):
Hey @UJESH2K, great catch on updating the TypeScript models to properly reflect that `tolerations` should be an array (`kubernetes.Toleration[]`) rather than a single object. 

However, I noticed a potential data-loss issue in `KubernetesYamlService.ts`:
```typescript
      runner: {
        ...manifest.spec.components?.runner,
        tolerations: [tolerations]
      }
```
By hard-wrapping the incoming `tolerations` argument in an array, this completely overwrites any pre-existing tolerations on the manifest. If a user tries to configure multiple tolerations, the UI will only ever save the most recent one. 

We should probably append it to the existing array instead:
`tolerations: [...(manifest.spec.components?.runner?.tolerations || []), tolerations]`

Let me know if the intention was strictly to support only a single toleration from the UI, but otherwise, we should probably preserve existing array elements!

UJESH2K:
Thanks for pointing this out! You're right—I missed this case while implementing the fix. I'll verify the behavior more thoroughly, update the implementation if needed to preserve existing tolerations, add the necessary regression test, and push the changes to this PR. Thanks for catching it!

PR #5430, opened 2026-02-02
Fix 504 timeout: Add K8s client caching + QPS increase + benchmarks

shovan-mondal:
### Proposed changes

Fixes #5079  (504 Gateway Timeouts).

This PR addresses a critical concurrency anti-pattern in the subscriber's `GetGenericK8sClient`, where a new `kubernetes.Clientset` was being initialized for every single request.

**The Issue:**

1. **Rate Limit Bypass:** Creating a new client for every request resets the token bucket, effectively bypassing client-side rate limits and allowing the subscriber to flood the API server.
2. **Socket Exhaustion:** In production environments, this causes high TCP connection churn (new TLS handshake per request), leading to latency accumulation and eventually 504 Gateway Timeouts.

**The Fix:**

* Implemented `sync.Once` to enforce a **Singleton pattern** for the Kubernetes client (reusing the TCP connection).
* Optimized `rest.Config` with **QPS=50** and **Burst=100** to handle concurrent UI requests without client-side throttling.
* Added `client_perf_test.go` to validate the performance improvement.

**Benchmark Results:**
I ran a parallel benchmark simulating **20 concurrent requests**.

| Scenario | Total Time | Avg/Request | Connections | Status |
| --- | --- | --- | --- | --- |
| **Current Bug** (New Client/Request) | 40.15ms | 2.00ms | Multiple (20) |  Connection Waste |
| **Naive Fix** (Reuse + Low QPS) | 7.51s | 375.61ms | Single (1) |  Throttled |
| **This PR** (Reuse + High QPS) | **38.55ms** | **1.92ms** | **Single (1)** |  Optimal |

## Types of changes

What types of changes does your code introduce to Litmus? Put an `x` in the boxes that apply

* [ ] New feature (non-breaking change which adds functionality)
* [x] Bugfix (non-breaking change which fixes an issue)
* [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
* [ ] Documentation Update (if none of the other choices applies)

## Checklist

Put an `x` in the boxes that apply. You can also fill these out after creating the PR. If you're unsure about any of them, don't hesitate to ask. We're here to help! This is simply a reminder of what we are going to look for before merging your code.

* [x] I have read the [CONTRIBUTING](https://github.com/litmuschaos/litmus/blob/master/CONTRIBUTING.md) doc
* [x] I have signed the commit for DCO to be passed.
* [x] Lint and unit tests pass locally with my changes
[...]

PR #5601, opened 2026-07-31
fix(auth): reject unverified email claims in the OIDC callback

naruto-lgtm:
Tracking issue: #5605

## Proposed changes

`OAuthCallback` decodes `email_verified` off the ID token into `claims.Verified` and then never looks at it. It sets both `Username` and `Email` to `claims.Email` and hands that to `LoginUser`, whose repository implementation inserts by username and, on a duplicate key, returns the *existing* record:

```go
_, err = r.Collection.InsertOne(context.Background(), user)
if err != nil {
	if mongo.IsDuplicateKeyError(err) {
		var result = entities.User{}
		findOneErr := r.Collection.FindOne(context.TODO(), bson.M{"username": user.Username}).Decode(&result)
		...
		return &result, nil
```

The handler then signs a session JWT for whatever user came back and redirects with it. So an id_token carrying someone else's address with `email_verified: false` logs the caller straight into that person's account and every project it owns. `/oauth/callback` is registered on the root engine with no middleware, so nothing upstream of this stands between the provider's claims and the session token.

Since `OIDC_ISSUER` became a free-form setting the server talks to whatever provider an operator points it at, and self-service signup on Keycloak, or an Auth0 database connection, hands out exactly that token shape until the address is confirmed. The bundled Dex connectors (google, github) both report verified emails, so default installs are unaffected. An absent `email` claim is the same defect from the other side: those logins all collapse onto one shared account with an empty username, which is why the check covers both.

The claim is validated where it is first treated as an identity rather than further down in `LoginUser`, since that path is also the local username/password login and has no notion of a provider assertion.

## Types of changes

## Checklist

## Dependency

None.

`gofmt`, `goimports` and `make backend-services-checks` are clean, `go.mod`/`go.sum` are untouched, and `go test ./...` in `chaoscenter/authentication` passes.

The new tests in `dex_auth_handler_test.go` stand up a small OIDC provider on `httptest` and drive the callback end to end, so they exercise the real `Exchange` plus `Verify` path rather than stubbing it. On master the three rejection cases return `308` with `jwtToken=victim-session-token` in the `Location` header; with the patch they return `401` and `GetSignedJWT` is never reached. `TestOAuthCallbackAcceptsVerifiedEmail` pins the happy path so the check cannot silently start rejecting valid logins.

I did not open a public issue for this one, since SECURITY.md asks that vulnerabilities go to the security contacts instead. Happy to file a tracking issue if you would rather have one linked here.

Left alone deliberately, tell me if you want either folded in: the OAuth `state` token from `GenerateOAuthJWT` carries only `exp` and is not bound to the browser session, so it does not actually stop login CSRF, and keying accounts on `email` rather than the provider's `sub` still means a mailbox change at the IdP silently detaches a user from their projects. Both are larger behaviour changes than this one.

### DIFF
--- a/chaoscenter/authentication/api/handlers/rest/dex_auth_handler.go
+++ b/chaoscenter/authentication/api/handlers/rest/dex_auth_handler.go
@@ -120,6 +120,13 @@ func OAuthCallback(userService services.ApplicationService) gin.HandlerFunc {
 			c.JSON(utils.ErrorStatusCodes[utils.ErrServerError], presenter.CreateErrorResponse(utils.ErrServerError))
 			return
 		}
+		// The email is what the account is keyed on below, so it is only usable
+		// once the provider states that it belongs to the subject.
+		if claims.Email == "" || !claims.Verified {
+			log.Errorf("OAuth Error: subject %s has no verified email claim", idToken.Subject)
+			c.JSON(utils.ErrorStatusCodes[utils.ErrUnauthorized], presenter.CreateErrorResponse(utils.ErrUnauthorized))
+			return
+		}
 		createdAt := time.Now().UnixMilli()
 
 		var userData = entities.User{
--- a/chaoscenter/authentication/api/handlers/rest/dex_auth_handler_test.go
+++ b/chaoscenter/authentication/api/handlers/rest/dex_auth_handler_test.go
@@ -1,11 +1,25 @@
 package rest_test
 
 import (
+	"crypto/rand"
+	"crypto/rsa"
+	"encoding/base64"
+	"encoding/json"
+	"math/big"
+	"net/http"
 	"net/http/httptest"
 	"testing"
+	"time"
 
+	"github.com/gin-gonic/gin"
+	"github.com/golang-jwt/jwt/v4"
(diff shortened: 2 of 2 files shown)


PriteshKiri:
Hey @naruto-lgtm 
Thanks for your contribution

Could you please create a corresponding issue for this and link it to the PR?

naruto-lgtm:
Filed #5605 with the summary, PoC and impact, and linked it here. Kept the report to the OIDC provider shapes that actually emit the unverified-email tokens so triage has the reproducer in one place.

PriteshKiri:
Hey @naruto-lgtm 
Could you please review the failed CI check?

naruto-lgtm:
Looked into it: the failure was the Trivy image scan, not the build itself. The branch predated #5430, which bumped x/crypto and x/net in the authentication module, so the scan was flagging the old versions baked into the image. Rebased onto master to pick up those bumps; go test ./... in chaoscenter/authentication is still green locally. The new workflow run needs an approval to start since I'm a first-time contributor here.

SarthakJain26 (approved):

ispeakc0de (approved):

PR #5621, opened 2026-09-20
fix(graphql-server): chaos-infra stuck INACTIVE after abrupt subscriber disconnect

### LINKED ISSUE #4789: Re-activating chaos-infra after `INACTIVE` state or disabling via UI or disconnecting via `litmusctl disconnect`
## Question

I have Litmus Operator running Litmus 3.5.0 in GKE. I would like to be able to deactivate a chaos-infrastructure and reactivate it on ad-hoc basis. Once I've installed the CRDs I can enable the infra.

However, I've found that disabling it once (either via the UI or using `litmusctl disconnect chaos-infra`), I can't create an environment with the same name as before - since the INFRA_ID and ACCESS_KEY is different even if the names are the same.

I thought it might be because the ids are managed separately in Mongo collections. So I decided to delete the CRDs from the chaos-infra in the cluster. This causes the chaos-infra to show up as `INACTIVE`. But when I re-apply the manifests again, it remains as `INACTIVE` and never turns `ACTIVE` again. 

I would like to be able to save costs by turning off chaos-infra resources, but I always need to start a chaos-infra from scratch if I decided to disable it/delete the resources. Is there any way to re-activate a chaos infra after deleting the chaos-infra manifests?

VarshaUN:
## Proposed changes

Fixes #4789

`InfraConnect` tracks each connected chaos-infra in an in memory map (`data_store.Store.ConnectedInfra`). An entry is only removed when the original connection's context hits `ctx.Done()`.

If a subscriber disconnects abruptly (CRDs deleted without graceful shutdown) and reconnects before the server notices the old connection is dead, the old map entry is still there. The new, already verified connection was being rejected with `CLUSTER ALREADY CONNECTED`, and the stale entry was never cleaned up. Since there's no heartbeat or idle timeout on this stream, the server may never detect the old connection died on its own, so the infra stays `INACTIVE` for good. That matches the report: re applying manifests never brings it back to `ACTIVE`.

Fix: when a new verified connection comes in for an infraID that already has a map entry, close and remove the stale entry and let the new connection take over, instead of rejecting it.

## Types of changes

## Checklist

## Dependency
- None.

this fix handles the new connection arrives case, but a subscriber that crashes and never reconnects at all would still leave a stale entry indefinitely (no heartbeat/timeout exists on this stream today). Happy to open a separate follow-up issue/PR for that if it's considered worth addressing, rather than scope creeping this one.


VarshaUN:
Hey @PriteshKiri  , The Trivy check is failing on two pre existing HIGH CVEs in `google.golang.org/grpc (v1.82.1)` unrelated to this change , this PR doesn't touch any dependencies. Let me know if this is already tracked or if there's something I should do here.

The fuzz-build check is also failing, unrelated to this PR . `proto: file "extensions/extension.proto` is already registered between `github.com/googleapis/gnostic` and `github.com/google/gnostic-models`. It's crashing on init before any test logic runs, and the log shows ~12.5% of fuzz targets failing the same way, so it looks like a repowide dependency conflict rather than something in this diff.

